PR #9, opened 2014-12-24
Failing test for symlinks

adgad:
/cc @matthew-andrews 

Leaving this here as a failing test for following symlinks in partials dir. It's fixed in my fork of express-handlebars, which I've put in a PR for:
    "express-handlebars": "adgad/express-handlebars#follow-symlinks",

https://github.com/ericf/express-handlebars/pull/98/files


adgad:
Okay, so the node-glob PR around following symlinks seems to have stalled. Given that the bower link workflow is quite important for us, I propose we operate off a fork of express-handlebars for now, with one of the following hacky solutions:

https://github.com/adgad/express-handlebars/compare/hacky-symlink-fix?expand=1
(I don't know enough about globs to know why that works, but it does)

OR rolling back node-glob to the v3.x branch (which had different symlink behaviour) as suggested here https://github.com/isaacs/node-glob/issues/134.

Thoughts?

matthew-andrews:
Actually the hacky solution makes sense based on isaacs comment 

> The solution is not "follow symlinked dirs" as suggested in the OP here. The correct behavior is "stop following symlinked dirs as part of a globstar match, **but still resolve their contents against subsequent parts of pattern**".

Personally, though, I don't think it's hacky enough.

https://github.com/Financial-Times/next-express/pull/14

PR #16, opened 2015-01-09
added gzipping

wheresrhys:
@matthew-andrews @commuterjoy 

the threshold at which to gzip defaults to 1kb, but is configurable. Any opinions?

### DIFF
--- a/main.js
+++ b/main.js
@@ -15,6 +15,7 @@ var paragraphs = require('./src/paragraphs');
 var removeImageTags = require('./src/remove-image-tags');
 var normalizeName = require('./src/normalize-name');
 var ifEquals = require('./src/if-equals');
+var compression = require('compression');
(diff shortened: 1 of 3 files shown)


commuterjoy:
I'd prefer to do this on the CDN. In other words - the application should server html, the CDN should optimise it's delivery.

https://docs.fastly.com/guides/caching/how-do-i-manually-enable-dynamic-gzipping

wheresrhys:
makes sense... getting a weird failure on a test here I don't relish wrestling with anyway

wheresrhys:
I've added here https://app.fastly.com/#configure/service/f8585BOxnGQDMbnkJoM1e/version/182
I think I configured the caching correctly. I'll leave it to you to check and activate the version @commuterjoy

PR #17, opened 2015-01-15
added if all helper

wheresrhys:
also a little housekeeping @matthew-andrews


matthew-andrews:
:+1: lovely.

could you update the `README.md` to mention `ifAll`?

PR #23, opened 2015-01-29
Adding barrier component

aintgoin2goa:


main.js
```diff
@@ -8,6 +8,8 @@ var express = require('express');
 var errorsHandler = require('express-errors-handler');
 var flags = require('next-feature-flags-client');
 var expressHandlebars = require('express-handlebars');
+var barriers = require('next-barrier-component')(flags);
```
matthew-andrews: flags here might not be what you think it is…

aintgoin2goa: it seems to be from looking at the code, but I'm not going to use it anyway, so I'll take it out...

matthew-andrews:
Seems sensible :+1:

PR #25, opened 2015-01-31
added template inheritance helpers

wheresrhys:


main.js
```diff
@@ -97,3 +103,10 @@ module.exports = function(options) {
 
 	return app;
 };
+
+
+
+
+
+
+
```
matthew-andrews: &thinsp;

&nbsp;

&thinsp;

&nbsp;

&thinsp;

&nbsp;

matthew-andrews:
beautiful work :+1:

wheresrhys:
I considered an npm postinstall script too, but couldn't convince myself it was sensible. For one thing, there'd be a risk of having header@x sass from next-ui-setup but header@y templates unless you force the postistall script to run bower install in the app root, and then... oh it's a pickle! Worth mulling over improvements still, along with trying to get more of make and gulp into next-build-tools

PR #28, opened 2015-01-31
added helper to detect if a block has any content assigned to it

wheresrhys:


wheresrhys:
hmmm - may change this to something like `{{#ifSomeContent var1 var2 var3 blocks=['blockname1', 'blockname2']}}` ... will be more flexible by far. Will be making the slightly flaky assumption (as now) that a block with some markup in it will actually output content. Could be a problem with flags

PR #30, opened 2015-02-03
SSL compatible image service endpoint

matthew-andrews:
Thank you @samgiles!


samgiles:
Ha, no problem :).  Just note, this is the Origin so it is not behind a CDN, you're hitting Amazon (Ireland)  directly on this.

matthew-andrews:
actually that's a very good point.  we should use next-geebe

PR #37, opened 2015-02-09
default layout is noop

wheresrhys:
@matthew-andrews


main.js
```diff
@@ -87,7 +87,7 @@ module.exports = function(options) {
 	app.engine('.html', expressHandlebars({
 		extname: '.html',
 		helpers: helpers,
-		defaultLayout: 'default',
+		defaultLayout: 'noop',
```
matthew-andrews: isn't this the same as `defaultLayout: false` ?

wheresrhys: good point. only saw this after merging darn

PR #47, opened 2015-02-18
Remove symlink hack

adgad:
/cc @wheresrhys @matthew-andrews 

Been fixed upstream in https://github.com/ericf/express-handlebars/commit/a91ae6b04ef8005e393f1a08cec73ad35b75caef


matthew-andrews:
you'd probably best set https://github.com/Financial-Times/next-express/blob/master/package.json#L14 to v1.2.1

PR #54, opened 2015-03-09
oupt appname in data attribute

wheresrhys:


matthew-andrews:
neat!

layouts/wrapper.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html class="no-js">
+<html class="no-js" data-next-app="{{__name}}">>
```
matthew-andrews: spot the bug @wheresrhys

wheresrhys: bum - sorry

PR #60, opened 2015-03-12
Don't instrument internal routes, e.g. __gtg, __help

matthew-andrews:
cc @commuterjoy 

why write readable, intuitive code when you could write a single line of regex?


commuterjoy:
:+1: yup. 

People have started to mis-use the __ for other purposes, so we need to also fix that, but later.

PR #61, opened 2015-03-12
sketch of fetch metrics

wheresrhys:
@matthew-andrews


matthew-andrews:
twas just thinking about this.  gonna need to put some thought into stripping out sensitive info, e.g. elastic search url

matthew-andrews:
oh you've done that with the service regexs.  nice

src/express/fetch-metrics.js
```diff
@@ -0,0 +1,87 @@
...
+									// do timing etc
+									return res
+								})
+								.catch(function (err) {
+									// do something maybe
+									throw err
+								})
+
+			} else {
+				// send error to sentry 'You're using a new service -
```
matthew-andrews: yes! :+1:

matthew-andrews: or maybe even

```js
var express = require('ft-next-express');

express({
  metrics: {
    'capi1-get': /^http:\/\/blahblah/
  }
});
```

[could mix-in some repeated defaults like capi in next-express]

matthew-andrews:
this ought to be fairly easy to test with stubs.  i don't mind trying to do that tomorrow

matthew-andrews:
i'm really excited about this.  it's such a neat solution to this.

src/express/fetch-metrics.js
wheresrhys: Yeah - I like the idea that sentry can be a peer reviewer. I wonder if we can hook IFTTT into it somehow?

matthew-andrews: for another thread perhaps :)

wheresrhys:
I'm leaning towards including most of this as a method in next-metrics (perhaps called within metrics.init() by defaut?) rather than directly in next-express. Although I think a `.registerCollector()` method for next-metrics would still be nice

README.md
```diff
@@ -22,8 +22,20 @@ For vanilla and wrapper layouts scripts and styles must still be manually requir
...
 - Provides a range of [handlebars helpers](#handlebars-helpers), including template inheritance
+- instruments `fetch` to send data about server-to-server requests to graphite. By default capi1, capi2, sapi and elastic search are instrumented. To add more services pass in a 'serviceDependencies' option (see examples below)
 
 
+var serviceMatchers = {
+	'capi-v1-article': /^https?:\/\/api\.ft\.com\/content\/items\/v1\/[\w\-]+/,
+	'capi-v1-page': /^https?:\/\/api\.ft\.com\/site\/v1\/pages\/[\w\-]+/,
+	'capi-v1-pages-list': /^https?:\/\/api\.ft\.com\/site\/v1\/pages/,
+	'sapi': /^https?:\/\/api\.ft\.com\/content\/search\/v1/,
+	'elastic-mget': /^https?:\/\/[\w\-]+\.foundcluster\.com:9243\/v1_api_v2\/item/,
```
matthew-andrews: not necessarily mgets

wheresrhys: Why? Because it's a temporary solution?

I see no problem with putting everything we use in 2 or more apps in here. Removing them from the list wouldn't be a breaking change. It's actually quite nice because if we intend to remove the service globally we can delete the listing here and any lingering use of the service will start sending messages to sentry.

matthew-andrews: no i mean that end point supports gets as well as mgets

matthew-andrews: suggest renaming it to `elastic-v1-article`

wheresrhys: I see

PR #62, opened 2015-03-13
exposing flags to css

wheresrhys:
To tackle css only bugs (like the current image service svg problem). Only exposing offy flags for the time being. 

@adgad @matthew-andrews

### DIFF
--- a/layouts/vanilla.html
+++ b/layouts/vanilla.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html class="no-js" data-next-app="{{__name}}">
+<html class="no-js" data-next-app="{{__name}}" {{{flagStatuses}}}>
 	<head>
 		<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1"/>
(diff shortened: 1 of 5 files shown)


ironsidevsquincy:
any benefit of data attrs over classes?

matthew-andrews:
precedent…  we use data attrs a lot…  i was uncomfortable at first but iirc they have the same specificity so doesn't really matter…

ironsidevsquincy:
just thinking about accessing them in older browsers

wheresrhys:
the `~=` selector works way back to ie7 (which surprised me when I found out). I prefer data attributes for on/offy things or things where the alternative is a long list of name-spaced classes. Neater and the intent is a bit clearer I think.

**By the way, no-one merge this yeat as it'll probably make it slightly harder to roll-out a quick fix for today's user-testing**

ironsidevsquincy:
wow, attribute selectors go down to ie7, really thought it would be higher than that. I take it all back, what a great browser

PR #64, opened 2015-03-18
Don't escape title and add class for o-hoverable

ironsidevsquincy:


layouts/vanilla.html
```diff
@@ -6,10 +6,10 @@
 		<meta name="theme-color" content="#fff1e0"/>
 		<meta name="dfp_site" content="test.next.ft" />
 		<link rel="stylesheet" href="/{{__name}}/main.css"/>
-		<title>{{#title}}{{this}} &mdash; {{/title}}Financial Times</title>
+		<title>{{#title}}{{{this}}} &mdash; {{/title}}Financial Times</title>
 		{{>next-mustard/check}}
 	</head>
-	<body>
+	<body class="o-hoverable-on">
```
matthew-andrews: is this best practise?  iirc the cutting edge was disable hover by default then if a mouse event detected add `:hover` rules back in…

[i realise i am guilty for not following this best practise [even today](https://github.com/Financial-Times/next-grumman/pull/258/files#diff-4b416eb6708da80a8f837e12882a266cR34)]

ironsidevsquincy: I think so, docs are a bit ambiguous. http://registry.origami.ft.com/components/o-hoverable@1.1.0 suggests you need the class first, and running the JS turns it off if appropriate. Doesn't seem to turn it on if the class isn't initially there

matthew-andrews:
:+1:

PR #69, opened 2015-03-25
Serve hashed assets from router

adgad:
/cc @wheresrhys @richard-still-ft @commuterjoy

### DIFF
--- a/src/handlebars/hashed-asset.js
+++ b/src/handlebars/hashed-asset.js
@@ -20,9 +20,8 @@ module.exports = function(options) {
 	} catch(err) {
 		console.warn("./public/asset-hashes.json not found.  Falling back to un-fingerprinted files.");
 	}
-
 	if (assetHash) {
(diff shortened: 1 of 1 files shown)


commuterjoy:
Not the comfortable with plugging the router in to non heroku domains. Would much rather we just have a assets.ft.com domain for the short term.

adgad:
Will do here for now then sort out a proper route for it.

adgad:
Proposed solution @commuterjoy @matthew-andrews @richard-still-ft :

* Add a basic express app to next-hashed assets. This will allow us to set cache headers etc.
* Push up to heroku (although has to be done synchronously before deploying app)
* Get next-geebee to point to app

matthew-andrews:
@commuterjoy 

> Not the comfortable with plugging the router in to non heroku domains. Would much rather we just have a assets.ft.com domain for the short term.

why?

PR #72, opened 2015-03-26
first steps towards better approach to bower deps

wheresrhys:
@matthew-andrews


wheresrhys:
See also https://github.com/Financial-Times/next-build-tools/pull/33

PR #76, opened 2015-03-31
Stub `console` object

ironsidevsquincy:
Sometimes ours, sometimes 3rd-party's code contain `console....` calls, which fail in IE<10

So stub it (and needs to be near the beginning of the chain)


ironsidevsquincy:
Actually, moving into `next-js-setup`

PR #77, opened 2015-04-01
Switch to npm version of next-metrics

matthew-andrews:


wheresrhys:
:+1: ... although I wonder if we should be running all tests with env=production, in which case the key is required.

matthew-andrews:
i don't think we should…  NODE_ENV should only be used for varying error escalation & analytics type behaviours — neither of which we would want automated tests to affect…  Any other uses of NODE_ENV if there are any possibly should be eliminated…

PR #82, opened 2015-04-07
Use next-logger

ironsidevsquincy:


package.json
```diff
@@ -10,10 +10,11 @@
     "es6-promise": "^2.0.1",
     "express": "^4.10.5",
     "express-errors-handler": "^1.0.0",
+    "ft-next-handlebars": "^1.0.1",
+    "ft-next-logger": "0.0.2",
```
matthew-andrews: I thought i bumped logger to v1.0.0

matthew-andrews: Mmmmmyess

ironsidevsquincy: yup, old PR, I'll update

PR #87, opened 2015-04-08
Add memwatch to log possible memory leaks

aintgoin2goa:
I think this would be a useful thing to have visibility of as we ramp up.

Note, this module is mostly written in C so the compilation step may slow the builds down

### DIFF
--- a/main.js
+++ b/main.js
@@ -11,6 +11,14 @@ var handlebars = require('ft-next-handlebars');
 var barriers = require('next-barrier-component');
 var metrics = require('next-metrics');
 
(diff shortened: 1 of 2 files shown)


matthew-andrews:
:+

matthew-andrews:
let's give it a try

matthew-andrews:
can we put it behind a flag?

matthew-andrews:
argh

https://github.com/lloyd/node-memwatch/issues/62

matthew-andrews:
https://app.getsentry.com/nextftcom/ft-next-grumman/group/61776009/

PR #89, opened 2015-04-08
Revert "Add memwatch to log possible memory leaks"

matthew-andrews:
Reverts Financial-Times/next-express#87


matthew-andrews:
also caused this: https://app.getsentry.com/nextftcom/ft-next-grumman/group/61776009/

:)

PR #95, opened 2015-04-17
Add session service middleware

aintgoin2goa:
Allows easy access to the session service to get user data

@matthew-andrews


main.js
```diff
@@ -99,6 +99,21 @@ module.exports = function(options) {
 		}
 	});
 
+	// Add in session helper
+	app.use(function(req, res, next){
+		res.locals.session = function(){
+			return fetch('https://session-next.ft.com/', {headers:{'Cookie: ' : req.get('Cookie')}})
```
matthew-andrews: i thought we stripped out cookies from fastly?

commuterjoy: We do

commuterjoy:
I think this is the wrong place for this. It would make the CDN (cache) redundant, the application layer never receives cookies.

What are you trying to do?

aintgoin2goa:
After talking it through I won't be adding this to next-express.  the service can now be called using a header rather than cookies, but we don't want to encourage devs to use user details in the back-end

PR #101, opened 2015-04-23
added session service

wheresrhys:


main.js
```diff
@@ -150,5 +151,5 @@ module.exports = function(options) {
 	return app;
 };
 
-module.exports.Router = express.Router
-
+module.exports.Router = express.Router;
+module.exports.services = serviceMatchers;
```
matthew-andrews: wai?

matthew-andrews: do you think it might be time for a separate `next-services` repo?

wheresrhys: we were talking about exposing the info so radiator can automatically build graphs for each service we're collecting for.

What else would go in next-services?

matthew-andrews: oh just thought you might want to have the array in a separate repo and then pull it in where you need it
but i'm fine with exposing it like this

PR #102, opened 2015-04-24
expose metrics

wheresrhys:
Don't want people requiring next-metrics directly as it will double up on the system performance stats @matthew-andrews

### DIFF
--- a/main.js
+++ b/main.js
@@ -10,7 +10,6 @@ var flags = require('next-feature-flags-client');
 var handlebars = require('ft-next-handlebars');
 var barriers = require('next-barrier-component');
 var metrics = require('next-metrics');
-var path = require('path');
 var robots = require('./src/express/robots');
 var normalizeName = require('./src/normalize-name');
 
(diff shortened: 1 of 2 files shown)


wheresrhys:
maybe it should throw rather than warn?

main.js
```diff
@@ -101,6 +101,12 @@ module.exports = function(options) {
 		});
 	}
 
+	if (require(path.join(process.cwd(), 'package.json')).dependencies['next-metrics']) {
```
matthew-andrews: hmm if we are going to do this we should do this everywhere……

matthew-andrews: oh actually if you're using `require` it uses `path.join` internally so it's not necessary

matthew-andrews: also i think `packageJson` is already in scope

main.js
```diff
@@ -101,6 +100,13 @@ module.exports = function(options) {
 		});
 	}
 
+	if (packageJson.dependencies['next-metrics']) {
```
matthew-andrews: can you just put the var packageJson a little higher up this file so that this won't syntax error if there's no package.json file?

matthew-andrews: before the try { 
`var packageJson = {};`

PR #117, opened 2015-05-13
Update the beacon url to use the .ft.com domain

matthew-andrews:
Should fix this: https://app.getsentry.com/nextftcom/ft-next-opt-in-out/group/65192069/

cc @commuterjoy


commuterjoy:
Thanks!

PR #118, opened 2015-05-14
Correct next-beacon host

commuterjoy:
should resolve https://app.getsentry.com/nextftcom/ft-next-opt-in-out/group/65192069/


matthew-andrews:
whoops

PR #119, opened 2015-05-14
Ability to set the base element

ironsidevsquincy:
cc @wheresrhys


wheresrhys:
:+1:, but maybe a test as well?

ironsidevsquincy:
now w/ added test

PR #122, opened 2015-05-21
Anon middleware

aintgoin2goa:


adamsilver:
Split up anon model and first click free model as discussed.

adamsilver:
:+1:

PR #125, opened 2015-05-26
exposes app version to client side

wheresrhys:
@matthew-andrews will use to configure o-errors


matthew-andrews:
:+1:

wheresrhys:
16 minor release in a row! When will the non-breaking goodness end??

PR #130, opened 2015-05-29
Dont throw http errors

matthew-andrews:
[better to just log them and let the app carry on as before]


matthew-andrews:
cc @aintgoin2goa

PR #138, opened 2015-06-08
Expose logged in state

adamsilver:


matthew-andrews:
can't we do this inside a js library? :/ :-1:

matthew-andrews:
also need to make sure that we only put this kind of data on pages that are varied on login status

matthew-andrews:
also also `nextData` is exactly the kind of global object we've tried before not to bring into Next.  (On FT.com they use `window.FT`)

aintgoin2goa:
@matthew-andrews all pages are varied on login status

I agree with the other stuff - though.  I thought this would just be a html-attribute?

adamsilver:
Just agreed with @matthew-andrews that we would expose a single variable nextIsAnonymous. Rationale for what it's worth: best to avoid a DOM read for performance.  Putting it all in data attributes doesn't solve the `window.FT` issues; you can see `documentElement` getting bigger and bigger :(

PR #139, opened 2015-06-09
Adding navigation items model

aintgoin2goa:


adamsilver:
Seems good besides failing build :)

PR #144, opened 2015-06-10
added internal capi to service list

wheresrhys:


matthew-andrews:
no!

matthew-andrews:
we should never use this in production

wheresrhys:
it is in production https://github.com/Financial-Times/next-url-mapper/blob/master/src/config.js - I did think it was a bit weird it was there

matthew-andrews:
ought to be fixed there then :wink:

wheresrhys:
`http://api.ft.com/site/v1/navigations/ft-uk` doesn't exist which is why this url is used.

matthew-andrews:
I emailed CAPI but ok for now

😔

PR #146, opened 2015-06-10
Add real first click free implementation

aintgoin2goa:
Note the new header names - these aren't (yet) set by the CDN

Also need to find out how to test this - not sure what the privileged referrers actually are...


adamsilver:
:+1:

PR #156, opened 2015-06-17
added bertha to service list

wheresrhys:


matthew-andrews:
:+1: maybe we should make a bertha spreadsheet of all the bertha spreadsheets we're using :wink:

PR #162, opened 2015-06-19
More barrier fixes

aintgoin2goa:
Am becoming very suspicious of the barrier API so adding some more logging and making sure it fails completely if the response is unexpected


adamsilver:
:+1:

PR #165, opened 2015-06-23
Added brand-alpha icons (and manifest for android).

adambraimbridge:


matthew-andrews:
chat with @richard-still-ft about geebeeing this urls :)

PR #167, opened 2015-06-23
remove internal capi

wheresrhys:
I'm not sure what the status of rolling out the new url is, so could you do the honours and merge this if/when it's gone from our apps @matthew-andrews


matthew-andrews:
it's gone :)

PR #169, opened 2015-06-24
Barrier beacon

aintgoin2goa:


matthew-andrews:
:+1:

matthew-andrews:
oh this was kind of a breaking change…… now apps won't start without `BEACON_URL`s…… annoying for APIs…… :/

ironsidevsquincy:
yup https://travis-ci.org/Financial-Times/next-dobi/builds/68185379

matthew-andrews:
not really a problem thoguh… those should probbaly set their `withBarriers` option to be false

matthew-andrews:
@ironsidevsquincy dobi can probably have `withBarriers: false` can't it @aintgoin2goa ?

matthew-andrews:
i'm going to do that for next-redirector

matthew-andrews:
grr still crashes

matthew-andrews:
@aintgoin2goa can we roll back this change for now until this: https://github.com/Financial-Times/next-beacon-node-client/issues/9 is fixed?

PR #178, opened 2015-06-29
Example sensu route

commuterjoy:
/cc @captain-sysadmin @matthew-andrews 

The idea here is each application own their own way of determining their health. By default this is just some canned response time metrics, status codes etc. 

Sensu can fetch the list of applications from http://next-registry.ft.com/ and hit the `__sensu` end point for each one to collect what we want it to monitor.

It allows us to standardise acceptable metrics across all applications - Eg, 500ms for every user-facing application server.

sensu

TODO

 - Allow applications to extend the __sensu endpoint.
 - Smoke test.
 - Extend to supporting the FT's __healthcheck standard.


src/express/sensu.js
```diff
@@ -0,0 +1,37 @@
...
+		},
+		{
+			check: util.format('sumSeries(%s)', responseTime),
+			name: "mean-response-time",
+			message: "Mean response time is above normal",
+			warn: 500,	// 500ms
+			critical: 1000,
+			occurences: 3,
+			interval: 60,
+			serviceLevel: 'gold',
```
matthew-andrews: gold?

commuterjoy: Was just playing with the idea that we can flag some metrics can be flagged as more important than others.

captain-sysadmin:
Yes! 

is it worth putting in a contact, or list of contacts in there as well? This would allow me to put a CC in the slack alert. (instead of @channel, which could be reserved for "gold" alerts)

Escalations would be good to put in numbers. After how many occurrences should it go to the next level of alerting? something like:

```
alert: slack_next_dev,
escalation: [
    pagerduty: 10
    taxi_service: 30
]
```

commuterjoy:
> is it worth putting in a contact, or list of contacts in there as well?

I will do once I've made a nice way to allow individual apps to add their own metrics.

> Escalations would be good to put in numbers

Perhaps we should start with pre-defined numbers?

PR #179, opened 2015-06-30
Added subscribe property based on anon state

adamsilver:


aintgoin2goa:
:+1:  (once build passes)

PR #181, opened 2015-06-30
update to match barrier api change and log request

aintgoin2goa:


commuterjoy:
Why are we building dependencies on legacy cookies?

aintgoin2goa:
The Barrier API requires this cookie as it is currently the ONLY way to tell if a customer is B2B or B2B.  literally.  This info is not stored anywhere else.

This cookie is still dropped by the login form as it cannot (yet) be replaced with anything else

Issue #155, opened 2015-06-17
a11y improvements


i-like-robots:
- [ ] Add off-screen anchors to navigation and main content

PR #186, opened 2015-07-03
Add accessibility skip links within wrapper template

phamann:
Adds hidden skip links to the top of the document enabling users using assistive technologies to jump to core sections of the page.

- Only enabled if `withHandlebars` and `withNavigation` are true.
- Requires extended applications to have a `u-visually-hidden` within their CSS
- ~~Extended applications cannot expose their own `#content` html id~~

Closes: #155

##### Potential improvements
- We could go further and follow GitHub's pattern by adding a `tab-index=1"` and visually exposing the element when focused?
- Use `<main>` element for the main content area?

/cc @i-like-robots @tom-parker

### DIFF
--- a/layouts/wrapper.html
+++ b/layouts/wrapper.html
@@ -44,9 +44,9 @@
 
 		{{>n-header-footer/templates/header}}
 
-		{{#if showSkipLinks}}<div id="content">{{#if}}
+		{{#if showSkipLinks}}<div id="content">{{/if}}
 		{{{body}}}
-		{{#if showSkipLinks}}</div>{{#if}}
+		{{#if showSkipLinks}}</div>{{/if}}
 
 		{{#if @root.flags.welcomePanel}}
 			{{>next-welcome/main}}
--- a/layouts/wrapper.html
+++ b/layouts/wrapper.html
@@ -36,7 +36,7 @@
 
 		{{#if showSkipLinks}}
 			<a class="u-visually-hidden" href="#primary-nav">Skip to navigation</a>
-			<a class="u-visually-hidden" href="#content">Skip to content</a>
+			<a class="u-visually-hidden" href="#site-content">Skip to content</a>
 		{{/if}}
 
 		{{#outputBlock 'notification'}}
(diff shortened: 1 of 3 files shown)


layouts/wrapper.html
```diff
@@ -33,12 +33,20 @@
 		{{/outputBlock}}
 	</head>
 	<body class="o-hoverable-on">
+
+		{{#if showSkipLinks}}
+			<a class="u-visually-hidden" href="#primary-nav">Skip to navigation</a>
+			<a class="u-visually-hidden" href="#content">Skip to content</a>
```
i-like-robots: I think apps are [currently using](https://github.com/search?q=site-content+user%3AFinancial-Times&type=Code&utf8=%E2%9C%93) the ID `site-content`

phamann: Nice! My only concern here is what if a new app forgets to add this. I.e. skip-link functionality with a broken anchor potentially worse to the user than not having one. It also makes it very hard to test.

i-like-robots: It should be easy to add a check to nightwatch and an item to any pre-launch checklists we have.

phamann: Have changed. But still don't like the fact we need to add to each nightwatch test suite and remember to add it. Where are these magical **pre-launch** check lists you speak of?

i-like-robots: It's not ideal, perhaps any app using next-express can inherit such a test somehow. 

I don't know if we have pre-launch checklists, it seems like the kind of thing we should have.

matthew-andrews:
if this went inside the header although it would be after the notification partial (not ideal) it would mean the Sass implementation would be a hell of a lot simpler

phamann:
@matthew-andrews Skip links need to be the first element within the page for them to be helpful for the user.

matthew-andrews:
just checking but can't see what we use that notification thing so might be able to delete it so that it would be

phamann:
Ok if so, would be glad to move the helper in there.

phamann:
Have moved too: https://github.com/Financial-Times/n-header-footer/pull/22

PR #191, opened 2015-07-06
Simple open graph tags

commuterjoy:
just the basics for now

### DIFF
--- a/layouts/wrapper.html
+++ b/layouts/wrapper.html
@@ -28,14 +28,14 @@
 		<meta name="msapplication-TileImage" content="https://next-geebee.ft.com/assets/brand-alpha/icons/mstile-144x144.png">
 		<meta name="theme-color" content="#fff1e0">
 		
-		{{#if @root.flags.openGraph}}
+		{{#if og}}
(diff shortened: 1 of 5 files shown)


layouts/wrapper.html
```diff
@@ -27,6 +27,16 @@
 		<meta name="msapplication-TileColor" content="#fff1e0">
 		<meta name="msapplication-TileImage" content="https://next-geebee.ft.com/assets/brand-alpha/icons/mstile-144x144.png">
 		<meta name="theme-color" content="#fff1e0">
+		
+		{{#if og}}
+		<meta property="og:title" content="{{ og.title }}"/>
+		<meta property="og:url" content="{{ og.url }}"/>
+		<meta property="og:description" content="{{ og.description }}"/>
+		<meta property="og:image" content="{{ og.image }}"/>
+		<meta property="og:type" content="article"/>
```
matthew-andrews: any reason why this stuff is here (particularly the `article` bit?  potentially all could live inside `next-article` making use of the head output block helper……?

commuterjoy: 'cos at some point I'll generate open-graph for all pages :)

PR #195, opened 2015-07-13
Allow applications to have their own sensu thresholds

commuterjoy:
/cc @captain-sysadmin @matthew-andrews


matthew-andrews:
lovely :)

PR #197, opened 2015-07-16
make warn 75% to reduce the spam in the sensu web interface

captain-sysadmin:
This ups the warn limit to 75%. Most users won't notice this as its only really obvious in http://sensu.ft.com


commuterjoy:
I've merged this locally.

PR #200, opened 2015-07-24
bind console methods to console

wheresrhys:
Useful for development

`.catch(console.log)` not `.catch(function (err) {console.log(err)})`


matthew-andrews:
😁 is this a good idea?

PR #201, opened 2015-07-27
Serve GET /__about from the root of the domain

matthew-andrews:


matthew-andrews:
To fit the FT standard

PR #203, opened 2015-07-29
added ability to pass in some metrics config when defining a route

wheresrhys:
Will need some work in next-metrics too, but the basic idea is

```
app.get('/article/:uuid', controller, {
   name: 'article',
   splitOn: [
     'extension',
     'queryString:fragment',
     'header:x-ft-anonymous'
   ]
}
```

where `splitOn` is an optional ordered list of additional things not covered by the route but which we may care about

If the config is missing (as it will be for all apps right now) then just defaults to using the route regex string as the name of the route

@commuterjoy @matthew-andrews

### DIFF
--- a/main.js
+++ b/main.js
@@ -66,6 +66,27 @@ module.exports = function(options) {
 		app.use('/' + name, express.static(directory + '/public'));
 	}
 
+	metrics.init({ app: name, flushEvery: 40000 });
+	app.use(function(req, res, next) {
+		metrics.instrument(req, { as: 'express.http.req' });
+		metrics.instrument(res, { as: 'express.http.res' });
+		next();
+	});
+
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -66,6 +66,27 @@ module.exports = function(options) {
...
+	metrics.init({ app: name, flushEvery: 40000 });
+	app.use(function(req, res, next) {
+		metrics.instrument(req, { as: 'express.http.req' });
+		metrics.instrument(res, { as: 'express.http.res' });
+		next();
+	});
+
+	serviceMetrics.init(options.serviceDependencies);
+
+	['get', 'put', 'delete', 'post'].forEach(function (method) {
```
matthew-andrews: head

PR #206, opened 2015-07-31
Align the graphite names for dependencies with their app equivalents

matthew-andrews:
@wheresrhys

### DIFF
--- a/src/service-metrics.js
+++ b/src/service-metrics.js
@@ -57,7 +57,10 @@ var serviceMatchers = {
 module.exports = {
 	init: function (additionalServices) {
 		if (additionalServices) {
-			Object.keys(additionalServices).forEach(function (serv) {
+			Object.keys(additionalServices).forEach(function(serv) {
+				if (serv.toLowerCase() !== serv || serv.indexOf('.') > -1) {
+					throw new Error("service dependency names should not have full stops in them and should be lowercase");
+				}
 				serviceMatchers[serv] = additionalServices[serv];
 			});
 		}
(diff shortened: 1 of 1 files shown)


matthew-andrews:
So we can draw pretty dependency graphs

src/service-metrics.js
```diff
@@ -49,15 +49,18 @@ var serviceMatchers = {
...
+	'ft-next-personalised-feed-api': /^http:\/\/(personalised-feed\.ft\.com|ft-next-personalised-feed-api\.herokuapp\.com)\/v1\/feed/,
 	'graphite': /^https:\/\/www\.hostedgraphite\.com\//
 };
 
 module.exports = {
 	init: function (additionalServices) {
 		if (additionalServices) {
-			Object.keys(additionalServices).forEach(function (serv) {
+			Object.keys(additionalServices).forEach(function(serv) {
+				if (serv.toLowerCase() !== serv || serv.indexOf('.') > -1) {
```
wheresrhys: I would also add a warning in here to say this functionality is deprecated and they should just add the service to next-express, then when we move all this to service-registry most of it will already be in one place

wheresrhys:
:+1:

src/service-metrics.js
matthew-andrews: is it deprecated?

wheresrhys: nope, but I think it should be.

matthew-andrews: ok, raise an issue with your reasoning.  i'm intrigued and don't have a strong opinion.

PR #209, opened 2015-08-04
Slightly more helpful error message?

adambraimbridge:


commuterjoy:
+1 

it's not really a whitelist - more a list of known hostnames.

PR #210, opened 2015-08-04
Backend access key

wheresrhys:


test/app.test.js
```diff
@@ -45,6 +45,58 @@ describe('simple app', function() {
...
+
+		xit('should 401 for arbitrary route with incorrect backend access key in production', function (done) {
+			request(app)
+				.get('/vanilla')
+				.set('FT-Next-Backend-Key', 'as-if')
+				.expect('ft-backend-authentication', /false/)
+				.expect(401, done);
+		});
+
+		it('should allow robots.txt through without backend access key', function (done) {
```
commuterjoy: why allow things to crawl robots?

wheresrhys: Um don't know. Could robots.txt prevent them crawling __about, __health etc... 

Happy to remove if this is nonsense

PR #211, opened 2015-08-04
Circuit breakers

phamann:
This PR is an Initial attempt at wrapping `fetch` requests within an application to external services (matching our serviceMatchers) with the [circuit breaker](http://martinfowler.com/bliki/CircuitBreaker.html) pattern.

For the time being this uses the Yammer [`circuit-breaker-js`](https://github.com/yammer/circuit-breaker-js) library as it does everything we needed out of the box, but could eventually abstract if we need in other locations and more custom functionality.
### Thoughts/questions
- What other metrics should we monitor
- ~~Should we put behind a flag or add a `withCircuitBreakers` prop.~~
- How can we properly test in production?
### TODO:
- [x] Add tests
- [ ] Try on a few more services
- [x] Add to healthcheck
- [x] Add alerting  

/cc @matthew-andrews

### DIFF
--- a/src/circuit-breakers.js
+++ b/src/circuit-breakers.js
@@ -22,10 +22,10 @@ Breaker.prototype.instrument = function(opts) {
 		this.serviceBreakers[service] = new CircuitBreaker(this.defaultBreakerOptions);
 
 		this.serviceBreakers[service].onCircuitOpen = function() {
-			opts.metrics.count('fetch_' + service + '_circuit_open', 1);
+			opts.metrics.count('fetch.' + service + '.circuit.open', 1);
 		};
 		this.serviceBreakers[service].onCircuitClose = function() {
-			opts.metrics.count('fetch_' + service + '_circuit_close', 1);
+			opts.metrics.count('fetch.' + service + '.circuit.close', 1);
 		};
 	}.bind(this));
 
--- a/src/circuit-breakers.js
+++ b/src/circuit-breakers.js
@@ -1,10 +1,11 @@
 'use strict';
 var CircuitBreaker = require('circuit-breaker-js');
+var Response = require('node-fetch').Response;
 
 var Breaker = function() {
 	this.serviceBreakers = {};
 	this.defaultBreakerOptions = {
-		windowDuration: 10000,
+		windowDuration: 30000,
 		numBuckets: 10,
 		timeoutDuration: 3000,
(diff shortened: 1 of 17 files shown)


matthew-andrews:
> Add alerting

You'll get this free with a healthcheck :)

matthew-andrews:
> Should we put behind a flag or add a withCircuitBreakers prop.

it doesn't fit well with the way we currently have flags architected.

to prototype maybe release this as a beta and put one or two apps on it?

src/circuit-breakers.js
```diff
@@ -0,0 +1,84 @@
...
+		this.serviceBreakers[service].onCircuitClose = function() {
+			opts.metrics.count('fetch_' + service + '_circuit_close', 1);
+		};
+	}.bind(this));
+
+	var _fetch = GLOBAL.fetch;
+	var that = this; // not using bind as don't want to muck around with fetch's scope
+
+	GLOBAL.fetch.prototype = Object.create(_fetch.prototype);
+	GLOBAL.fetch.constructor = function(url, opts) {
```
phamann: This bit of yucky inheritance was needed as it seems [`next-metrics`](https://github.com/Financial-Times/next-metrics/blob/master/lib/metrics/fetch.js#L73) places properties on the object which our tests depend on.

phamann: I'll probably use the same pattern in `next-metrics` too.

src/circuit-breakers.js
```diff
@@ -0,0 +1,89 @@
+/* jshint forin: false */
+'use strict';
+var CircuitBreaker = require('circuit-breaker-js');
+var Response = require('node-fetch').Response;
+
+var Breaker = function() {
+	this.serviceBreakers = {};
+	this.defaultBreakerOptions = {
+		windowDuration: 30000,
```
commuterjoy: why these defaults?

phamann: As @matthew-andrews and I decided 30s was a nice window ¯\_(ツ)_/¯

The point being, I'm 98% sure these defaults will need tweaking (and most likely at a service-by-service level). Which, is why we're going to test on one service first, view the metrics of how often it tripped within a week and adjust.

phamann:
I *will* ship this this week!

matthew-andrews:
i'll see it when i believe it

PR #215, opened 2015-08-07
Update manifest

adgad:
For all the push

/cc @matthew-andrews @phamann


matthew-andrews:
That has a massive max age.  You wanna hash?

matthew-andrews:
oh v1, ok.

matthew-andrews:
:+1:

PR #220, opened 2015-08-11
Remove x-powered-by express header

phamann:
Removes the: `x-powered-by=express` header. This is an unnecessary header causing more bytes over the wire and most importantly opens up to any potential express based security exploitations in the future.

http://expressjs.com/api.html#app.settings.table

### DIFF
--- a/main.js
+++ b/main.js
@@ -61,6 +61,9 @@ module.exports = function(options) {
 	var sensuChecks = sensu(name, options.sensuChecks);
 	var healthChecks = options.healthChecks;
 
+	//Remove x-powered-by header
+	app.set('x-powered-by', false);
+
 	try {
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -61,6 +61,9 @@ module.exports = function(options) {
 	var sensuChecks = sensu(name, options.sensuChecks);
 	var healthChecks = options.healthChecks;
 
+	//Remove x-powered-by header
+	app.set('x-powered-by', false);
```
commuterjoy: surely, `app.set('x-powered-by', 'biscuits');`

matthew-andrews:
:+1: every byte is sacred, every byte is good, if a byte is wasted …

PR #222, opened 2015-08-11
Circuit break with health

matthew-andrews:
@phamann


matthew-andrews:
Kind of hackily done for now…… also merges from master [sorry :see_no_evil:]

PR #223, opened 2015-08-11
Circuit breaker with health

matthew-andrews:


matthew-andrews:
[with a few commits cherry picked off of master]

test/circuit-breakers.test.js
```diff
@@ -40,34 +39,51 @@ describe('circuit breakers', function(){
...
 			breaker.volumeThreshold = 1;
 
+			return app.listen(4242);
 		});
 
-		it('should trip circuit if service fails', function(done){
-			fetch('http://localhost:3000/__fail')
+		afterEach(function() {
+			// HACK: Reset the breaker
+			breaker._buckets = [breaker._createBucket()];
```
matthew-andrews: this should probably simulate the upstream API getting healthy again…… but i couldn't figure out how to make it recover…… probably because i'm tired but maybe a bug……

matthew-andrews:
nb. this is to merge into your branch, @phamann

test/circuit-breakers.test.js
matthew-andrews: possibly because of the fact we're running two apps …… there might be multilpe circuit breakery interfering with each other…… might want to move the code that sets up the circuit breaker for this app into the main fixture app so that there's less chance of things interfering with each other……

PR #228, opened 2015-08-19
reinstating sticky footer as an option

wheresrhys:
I figure it's nice to be able to turn this on for things like error pages, confirmation pages, where janky scroll is not so much of an issue @matthew-andrews


matthew-andrews:
:+1:

PR #232, opened 2015-08-22
Also support IP address for portfolio

matthew-andrews:


matthew-andrews:
Fixes much warning https://app.getsentry.com/nextftcom/ft-next-ilyushin/group/77210250/

PR #235, opened 2015-08-25
Add preconnect hints to wrapper template

phamann:
This adds new `preconnect` hints (part of [resource hints API](http://w3c.github.io/resource-hints/#preconnect)) to give the browser hints to which hosts are critical for our application and we would like create a TCP connections for in advance before the parser finds the resources in the DOM. Ultimately reducing page load time.

To start with I've used the FT hosts which are found of every page. But have also allowed the ability for applications to add their own such as Livefyre for article pages etc.

/cc @richard-still-ft @matthew-andrews @aintgoin2goa

### DIFF
--- a/layouts/wrapper.html
+++ b/layouts/wrapper.html
@@ -8,6 +8,13 @@
 		{{#if setBase}}
 			<base target="_parent" href="{{setBase}}"/>
 		{{/if}}
+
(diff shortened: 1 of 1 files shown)


matthew-andrews:
:+1:

PR #238, opened 2015-08-26
Remove layouts

i-like-robots:
Layouts to be provided by https://github.com/Financial-Times/n-layout which should reduce some of our dependency spaghetti and allow us to write styles for layouts.

One of the bits of spaghetti this solves is for the next-welcome component which is currently used like this:

1. HTML is included here: https://github.com/Financial-Times/next-express/blob/4ba0f1dbb33d6050734b505f78ca8ded52d2ea8b/layouts/wrapper.html#L61
2. JS is instantiated here: https://github.com/Financial-Times/next-js-setup/blob/bc6a8b68d11da53a655c50cf7f58719c3031d99d/src/js-setup.js#L101-L105
3. CSS is included here, here and here: https://github.com/search?l=scss&q=next-welcome+user%3AFinancial-Times&type=Code&utf8=%E2%9C%93


main.js
```diff
@@ -148,7 +148,7 @@ module.exports = function(options) {
 				directory + '/views/partials'
 			],
 			defaultLayout: false,
-			layoutsDir: options.layoutsDir || __dirname + '/layouts',
+			layoutsDir: options.layoutsDir || process.cwd() + '/client/layouts',
```
wheresrhys: shouldn't this point to something like `process.cwd() + '/bower_components/n-layout/layouts'`

i-like-robots: Because n-layout isn't a dependency of next-express I didn't want to make such a specific assumption.

wheresrhys: But then you force every app to copy the layouts across to client, or manually set the layouts directory. I think the default should definitely be to point to the bower component. `nbt verify` can be tweaked to break builds if n-layout is not installed

i-like-robots: We've built APIs on top of this that shouldn't have any client-side deps though?

wheresrhys: `nbt verify --skip-layout-checks` already exists for this eventuality

wheresrhys: or, um, cleverer 
```
if (bower_components/n-layout exists) {
    layoutsDir = options.layoutsDir || process.cwd() + '/bower_components/n-layout/layouts'
} else {
   layoutsDir = options.layoutsDir || null
}
```

matthew-andrews:
:+1: any pull request that's overwhelming red!

wheresrhys:
I still strongly think the default layouts dir should be `/bower_components/n-layout/layouts`.

PR #246, opened 2015-09-16
Healthcheck promises

Lrnk:
@matthew-andrews I've fiddled the health check stuff in next express so it can handle async (like we need for the myft api health check)

Not sure I'm pleased with how I'm handling the case where the health check results in a failing promise (which the docs advise should be avoided). At the minute it just 500s, as I can't think of a reasonable way to make it more precise. Any ideas?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 /node_modules/
 /bower_components/
+.idea/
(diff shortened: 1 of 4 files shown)


matthew-andrews:
hmm…  all the healthchecks we currently have are not using asyncrhonous logic… they sit healthchecking (polling, etc) and then the health check endpoint outputs the current state synchronously.

it's fine/good for healthchecks can be unhealthy when apps start

matthew-andrews:
sorry… i don't think health checks should be allowed to be asynchronous :-1:

Lrnk:
Ah yep, you're absolutely right, that's what the lastUpdated flag is for I suppose. Please disregard.

PR #251, opened 2015-09-21
Bump next-navigation to 4.0

charypar:
This should be a change without any side-effects


matthew-andrews:
if you're happy that this won't break things :ship:

PR #259, opened 2015-09-29
Export the express.static class method to use within applicaitons

phamann:
I have the need to use the `express.static` middleware within an application to serve some precompiled documentation. We currently don't expose any of Expresses built in methods such as `static` to the outer application. 

### Usage:
`app.use('/docs', express.static('public/docs'));`

### DIFF
--- a/main.js
+++ b/main.js
@@ -222,6 +222,7 @@ module.exports = function(options) {
 };
 
 module.exports.Router = express.Router;
+module.exports.static = express.static;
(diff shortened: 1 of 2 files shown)


matthew-andrews:
ft next express should feel like normal express so we should amble towards exporting the same interface.  i say amble because i think we should only do that when we actually need/use it.

matthew-andrews:
:+1:

PR #265, opened 2015-10-08
[DO NOT MERGE] Added Yahoo's CSP module; Set up CSP reporting for next.ft.com

adambraimbridge:
cc @commuterjoy @matthew-andrews 

I'm not 100% sure about the logging portion of this PR (main.js L92) because I don't know how to test it locally. Any advice welcome.

### DIFF
--- a/main.js
+++ b/main.js
@@ -90,7 +90,7 @@ module.exports = function(options) {
 		});
 
 		// Log CSP violation reports to splunk via nextLogger
-		app.get('/__csp-reports', function(req, res) {
+		app.get('/__csp-reports', function(req) {
 			if(app.locals.__isProduction) {
 				nextLogger.logger.warn(JSON.stringify(req.body));
 			}
--- a/main.js
+++ b/main.js
@@ -4,6 +4,7 @@
 require('isomorphic-fetch');
 
 var express = require('express');
+var csp = require('express-csp');
 var errorsHandler = require('express-errors-handler');
 var flags = require('next-feature-flags-client');
 var handlebars = require('@financial-times/n-handlebars');
@@ -28,6 +29,7 @@ module.exports = function(options) {
 		withNavigation: true,
 		withAnonMiddleware: true,
 		withBackendAuthentication: true,
+		withContentSecurityPolicy: true,
 		sensuChecks: [],
 		healthChecks: []
 	};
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -68,6 +70,33 @@ module.exports = function(options) {
...
+						'*.ft.com',
+						'*.fyre.co',
+						'*.livefyre.com'
+					]
+				}
+			}
+		});
+
+		// Log CSP violation reports to splunk via nextLogger
+		app.get('/__csp-reports', function(req) {
```
matthew-andrews: hmm i don't think this is going to work how you think

matthew-andrews: i think we should make a new service on `/__csp-reports` for collecting CSP reports and passing them to Splunk rather than trying to put it into next-express.

matthew-andrews: `next-csp-ingest` ?

adambraimbridge: Hm ... would a more generic logging endpoint be ultimately more useful? Like `next-splunk-ingest`? Do we already have a generic logging endpoint for next?

matthew-andrews: i think we rarely ingest from the client side…  although as these are error maybe you could use the sentry client and log them to sentry instead of splunk?

adambraimbridge: I spoke with Matt A — I'll make an app that utilises next-logger and host it on heroku

adambraimbridge:
I've created this to ingest CSP reports: https://github.com/Financial-Times/next-csp-ingest
• it just inits o-tracking and fires a page() event. 
• it's hosted here: https://ft-next-csp-ingest.herokuapp.com/__csp-ingest

Note: This spews out a red river of console logs ... not sure how to intelligently disable that.
image

main.js
```diff
@@ -68,6 +70,26 @@ module.exports = function(options) {
 	var sensuChecks = sensu(name, options.sensuChecks);
 	var healthChecks = options.healthChecks;
 
+	// Note: Only *report* CSP violations.
+	// The aim is to know which scripts are being executed via third-parties.
+	if (options.withContentSecurityPolicy) {
+		csp.extend(app, {
+			reportPolicy: {
+				directives: {
+					'report-uri': ['https://ft-next-csp-ingest.herokuapp.com/__csp-ingest'],
```
matthew-andrews: ahh please use .ft.com URLs (either this should be sent through the router on `next.ft.com` or should have its own domain (see konstructor.svc.ft.com to set up DNS and Fastly.com to do the SSL termination) …… sorry :see_no_evil:

adambraimbridge:
Closing this for now.

Instead of setting CSP report headers in next-express, they're set in Fastly via VCL. See: http://git.svc.ft.com/projects/NEXT/repos/fastly-config/pull-requests/98/overview

In future, if we want to be able to set whitelisted third-party URLs on a per-feature basis, we'll need to re-open this PR, or more likely move it to `next-js-setup`, which I believe is the earliest point we have access to the flags API

main.js
adambraimbridge: No worries @matthew-andrews ... we're going to run this in reporting mode for a week or two just as an experiment, and if we decide to go ahead I'll sort out the csp-ingest URL.

PR #268, opened 2015-10-12
agnostic service protocols

wheresrhys:


matthew-andrews:
oh… i was using this to enforce https :/

wheresrhys:
oh. It seemed a bit haphazard... sorry didn't pick up on the pattern. I'll roll back

PR #273, opened 2015-11-03
[temporarily] disable geebee for js and css

matthew-andrews:
Just to rule out Akamai/geebee being the cause of this: https://app.getsentry.com/nextftcom/ft-next-frontend/group/89538895/


adambraimbridge:
:+1:  I'm going to see if I can find stats on non-2xx responses for https://next-geebee.ft.com/*

PR #275, opened 2015-11-04
Fixes an app that proxies to myFT

keirog:
We're bringing back an absolute URL in order to fix access to assets via the [next-clippings-proxy](https://github.com/Financial-Times/next-clippings-proxy) app, which proxies to the myFT tour page. (Not using next-geebee while a separate issue is being diagnosed.)


phamann:
:+1:

wheresrhys:
ah, this may fix next-flags too

PR #279, opened 2015-11-10
Revert to geebee

richard-still-ft:
Akamai have pushed out a fix that should resolve the issues we were seeing


phamann:
:+1: :shipit:

PR #280, opened 2015-11-10
csrf protection

laurieboyes:
Security FTW: https://github.com/pillarjs/understanding-csrf

This is necessary for Next as the core experience for myFT requires requests to be made via forms.

Using [csurf](https://www.npmjs.com/package/csurf), this PR puts a value `csrfToken` in `res.locals` (which then appears on the Handlebars root context) that needs to be provided with any form or XHR request that isn't a GET, HEAD or OPTIONS.

This'll be a breaking change for any app that makes a request like this so will have to be a major version bump.

### DIFF
--- a/main.js
+++ b/main.js
@@ -16,6 +16,9 @@ var normalizeName = require('./src/normalize-name');
 var anon = require('./src/anon');
 var serviceMetrics = require('./src/service-metrics');
 var dependencies = require('./src/dependencies');
+var cookieParser = require('cookie-parser');
+var bodyParser = require('body-parser');
+var csurf = require('csurf');
 
 module.exports = function(options) {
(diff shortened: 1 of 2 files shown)


wheresrhys:
Not sure I 100% get this, but is the plan to make some changes in fastly too? all cookies get stripped before the request gets to express

laurieboyes:
Ahhh I hadn't realised that. Cheers, I'll see if there's a way around it

laurieboyes:
This ain't happening. Cookie and session issues aside,  doing this will generate *uncachable pages*, which is very very silly. Further discussion has lead to the realisation that the worst any assailant can do with CSRF is force someone to follow something or delete their lists or whatever.

PR #282, opened 2015-11-18
Always show the myFT link in the main nav

keirog:
No longer limited to signed-in users.


keirog:
Got an IRL :+1: from @wheresrhys

PR #284, opened 2015-11-26
middleware to set global vary headers

wheresrhys:


wheresrhys:
Maybe this is overkill, but it means it's impossible to accidentally unvary by overwriting the whole vary header using `res.set()`. Can't build it around the existing `res.vary()` method as it calls `res.set()` internally, so end up with unwanted recursion

PR #285, opened 2015-12-10
Add editions locals

ironsidevsquincy:
Add current and 'other' editions as locals


matthew-andrews:
possibly better for this to be in n-navigation :/

matthew-andrews:
https://github.com/Financial-Times/n-navigation

ironsidevsquincy:
hmmmm...possibly. not sure if 'navigation' is the right umbrella term for it. do agree that it should be moved somewhere though

PR #289, opened 2015-12-22
product-selector -> products

commuterjoy:


matthew-andrews:
:+1: for the change, but what's this doing here?  can it be moved away in your refactor too?

commuterjoy:
> but what's this doing here

It looks like it's flagging anonymous users to the controllers and templates.

I might be able to remove it, but applications can serve anonymous people - need to investigate.

wheresrhys:
still needed @commuterjoy ?

commuterjoy:
nah.

PR #292, opened 2016-01-13
Add subscribe and sign-in URLs to the navigation model

keirog:
- moves navigation model tests out into their own file
- adds subscribeUrl and signInUrl to avoid duplicating links all over the site


keirog:
Might be better in config vars…

matthew-andrews:
config vars only required for secrets  / stuff that need to change separate from deploys… this looks less essential to be there.  so basically, it's your call.

keirog:
Ok ta. Probably easier to maintain it here so will merge.

PR #295, opened 2016-02-11
Use latest n-logger

ironsidevsquincy:


matthew-andrews:
n all the things :+1:

PR #310, opened 2016-04-01
V16

matthew-andrews:
- [x] Forces an `n-makefile.json` file to be defined and contains a mapping from unbuilt to build CSS + JS files (used by both `n-express` and `n-makefile`).
- [x] Removes `/__sensu` integration.
- [x] Rename to `@financial-times/n-express`
- [x] `n-makefile`-ify
- [x] Removes ijento logic
- [x] Remove broken `/__dependencies` logic
- [x] Remove a bunch of unused dependencies (including next build tools :scream:!!!)
- [x] Switches the default options so that `n-express` defaults to most things off (flags, etc) - TBD
- [x] Removes the (deprecated) ability to set custom per app service metrics


main.js
```diff
@@ -236,5 +224,5 @@ module.exports.static = express.static;
 module.exports.services = metrics.services;
 module.exports.metrics = metrics;
 module.exports.flags = flags;
-// TODO: for backwards compatiability, but modules/apps can use next-logger directly, as of v4.0.0
-module.exports.logger = nextLogger;
+// TODO: for backwards compatiability, but modules/apps can use n-logger directly, as of v4.0.0
+module.exports.logger = nLogger;
```
wheresrhys: can we delete this. It's led to problems before. There's a good reason to expose flags and metrics as we only want a single instance of those running in any app, but logger can be instantiated multiple times in an app AFAIK

cc @ironsidevsquincy

main.js
```diff
@@ -236,5 +224,5 @@ module.exports.static = express.static;
 module.exports.services = metrics.services;
```
wheresrhys: I wonder if we need this?

wheresrhys:
Do we have a replacement for the dependencies thing? What was broken about it?

main.js
ironsidevsquincy: n-logger is only instantiated once anyway (https://github.com/Financial-Times/n-logger/blob/master/src/main.js#L80-L89), initially it was in here so it could get the name of the app that was running, to construct the splunk url. now that's a config-var, there's no reason to use this. however, there might still be apps using it this way. 

in summary, this should be removed, just need to do a sweep of the apps to make sure nothing's using it

matthew-andrews:
@wheresrhys it was just broken 500-ing on all apps, i think graphite dashboard has the same data anyway

main.js
matthew-andrews: ok i think this is a good opportunity to remove it, i will do that

wheresrhys:
oh, it's not the code I thought it was.

main.js
matthew-andrews: i'll delete it

wheresrhys:
One last request, delete lines 163 to 170 in main.js

PR #312, opened 2016-04-04
Vary on correct anonymous user header

phamann:
As is causing caching issues :/


matthew-andrews:
how did that happen?

wheresrhys:
:+1:

phamann:
Well strangely it seems we are still setting both: https://github.com/search?q=org%3AFinancial-Times+FT-Anonymous-User&type=Code

I will continue to investigate as to where'/why it's being removed. But want to ship fix in the interim.

matthew-andrews:
+1

matthew-andrews:
fyi this is a new major version…… might want to patch v15

PR #320, opened 2016-04-14
support tracing in ft-next-express@15

wheresrhys:
@phamann @ironsidevsquincy

### DIFF
--- a/main.js
+++ b/main.js
@@ -62,7 +62,7 @@ module.exports = function(options) {
 			process.env.TRACE_SERVICE_NAME = normalizeName(name);
 			require('@risingstack/trace');
 		} else {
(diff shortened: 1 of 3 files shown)


matthew-andrews:
aw, why can't upgrade people to 16?

wheresrhys:
Because it's not my job to upgrade all the apps. It's my job to collect meaningful request timings ASAP.

Original plan was to upgrade everything/encourage others to but as n-makefile isn't on v1 yet and my experience upgrading just 2 apps wasn't 100% smooth I think that task should be distributed around the team once we get to v1. Maybe start failing builds for using express 15 a week after n-makefile gets to v1

matthew-andrews:
n-express 16 isn't dependent on n-makefile

PR #323, opened 2016-04-18
stop varying on country code for everything

wheresrhys:
@matthew-andrews @ironsidevsquincy 

cc @commuterjoy @ifyio @lc512k Are you explicitly setting vary on country-code headers everywhere you need to in barriers/product-selector etc. After this is merged, by default apps won't vary on country-code

### DIFF
--- a/src/middleware/vary.js
+++ b/src/middleware/vary.js
@@ -10,7 +10,7 @@ function extendVary (val, set) {
 
 module.exports = function(req, res, next) {
 	const resSet = res.set;
-	const varyOn = new Set(['country-code']);
+	const varyOn = new Set([]);
 
 	res.set('vary', Array.from(varyOn).join(', '));
 
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -65,7 +65,7 @@ describe('simple app', function() {
 		it('set default vary headers', function (done) {
 			request(app)
 				.get('/default-vary')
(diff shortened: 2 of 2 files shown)


matthew-andrews:
i think the sign up app needs this adding

wheresrhys:
It uses strict no-cache on everything so I think we're safe there

matthew-andrews:
oh, wow, that's odd.

wheresrhys:
I started a branch to apply vary country-code, but will discuss further with anon team. In the meantime I think it's safe to merge this

commuterjoy:
articles + barriers will need to vary on country-code as they share a URL

matthew-andrews:
oh really?  might be worth adding an item to our performance backlog to look into seeing if we can make that better.

wheresrhys:
what's the advantage of serving barriers on the same url?

matthew-andrews:
none afaik

matthew-andrews:
would make things simpler if it weren't^^

wheresrhys:
I suppose with the same url you get

fastly -> preflight -> barrier

whereas with different url you get

fastly -> preflight -> redirect -> fastly -> preflight -> barrier

wheresrhys:
https://github.com/Financial-Times/next-article/pull/1316

PR #325, opened 2016-04-20
Rhys/cache presets

wheresrhys:
API:

```
// 5 presets
res.cache('no');
res.cache('short');
res.cache('hour');
res.cache('day');
res.cache('long');

// use a preset but override bits of it
res.cache('day', {
  staleIfError: 20000000000000000,
  public: true
});

// full control, which still prevents you doing really stupid things
res.cache(surrogateControlHeaderString, cacheControlHeaderString) doing really stupid things
```

@phamann @matthew-andrews @ironsidevsquincy


src/middleware/cache.js
```diff
@@ -0,0 +1,121 @@
+'use strict';
+
+// Note - private/public declarations will automatically be added as appropriate if not specified
+const presets = {
+	no: {
+		maxAge: 0,
+		conditions: 'no-cache, no-store, must-revalidate'
```
matthew-andrews: shouldn't it also have `private`?

matthew-andrews: i think fastly only looks at private otherwise it'll cache…

wheresrhys: There's an internal method `privatize()` used to add private. I figured it's easier to define everything without public/private and add one or the other as a final step rather than doing string replaces. Or something... can't quite remember the rationale, or whether I half evolved away from it, now

src/middleware/cache.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+const presetStrings = Object.keys(presets).reduce((obj, key) => {
+	obj[key] = privatize(headerObjToString(presets[key]));
+}, {});
+
+const presetNames = Object.keys(presets).join('\n');
+
+function isHeaderString (str) {
+
```
ironsidevsquincy: empty?

wheresrhys: ha, yes. very much a WIP.

src/middleware/cache.js
wheresrhys: Ah I misunderstood you there. This is also used to set the outbound cache header. Perhaps should be stored in a different variable though

PR #332, opened 2016-05-04
Rhys/better anon detection

wheresrhys:
@lc512k @laurieboyes @ifyio


src/anon/models.js
```diff
@@ -1,12 +1,12 @@
 'use strict';
 
 function AnonymousModel(req){
-	if(req.get('FT-User-UUID')){
-		this.userIsLoggedIn = true;
-		this.userIsAnonymous = false;
-	}else{
+	if (req.get('FT-Anonymous-User') === 'true'){
```
wheresrhys: This is the only real change here - the rest is linting

PR #334, opened 2016-05-05
Add `hasHeadCss` option

ironsidevsquincy:
To read the head css file and store in `res.locals.headCss`

Addresses https://github.com/Financial-Times/next-front-page/pull/758#discussion-diff-62103412 - @matthew-andrews

### DIFF
--- a/main.js
+++ b/main.js
@@ -3,6 +3,7 @@
 
 require('isomorphic-fetch');
 
+const denodeify = require('denodeify');
 const express = require('express');
 const raven = require('@financial-times/n-raven');
 const flags = require('next-feature-flags-client');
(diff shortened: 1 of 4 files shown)


main.js
```diff
@@ -28,6 +30,7 @@ module.exports = function(options) {
 		withAnonMiddleware: false,
 		withBackendAuthentication: false,
 		withRequestTracing: false,
+		hasHeadCss: false,
```
matthew-andrews: i assume that we are currently requiring a one-to-one mapping between apps and head css's.  ie. an app can't have more than one compiled head css file?  (e.g. different routes)

(this is ok, keeps things simple, just checking)

matthew-andrews:
+1

main.js
ironsidevsquincy: Correct, though good point, might be something we need in the future. Though if an app has wildly different heads between pages, maybe it shouldn't be within a single app...

matthew-andrews: yeah i agree with that.  quite nice to have an incentive to split things out for a change :]

PR #337, opened 2016-05-16
Add a warning in non production when assets may not have been built

i-like-robots:


wheresrhys:
:+1:

wheresrhys:
implementation now completely changed

PR #342, opened 2016-06-07
feat(security): allow use of older backend keys

keithamus:
This commit allows the use of up to 3 backend keys, which allows a grace period for older keys
to be rotated. However, if the environment variables for each key are not set, they will not be used

/cc @matthew-andrews

### DIFF
--- a/main.js
+++ b/main.js
@@ -18,6 +18,16 @@ const serviceMetrics = require('./src/service-metrics');
 const vary = require('./src/middleware/vary');
 const cache = require('./src/middleware/cache');
 const headCssMiddleware = require('./src/middleware/head-css');
+const backendKeys = [];
+if (process.env.FT_NEXT_BACKEND_KEY) {
+	backendKeys.push(process.env.FT_NEXT_BACKEND_KEY);
+}
+if (process.env.FT_NEXT_BACKEND_KEY_OLD) {
+	backendKeys.push(process.env.FT_NEXT_BACKEND_KEY_OLD);
+}
+if (process.env.FT_NEXT_BACKEND_KEY_OLDEST) {
+	backendKeys.push(process.env.FT_NEXT_BACKEND_KEY_OLDEST);
+}
 
 module.exports = function(options) {
 
@@ -90,7 +100,7 @@ module.exports = function(options) {
 				// allow healthchecks etc. through
 				req.path.indexOf('/__') === 0) {
 				next();
-			} else if (req.get('FT-Next-Backend-Key') === process.env.FT_NEXT_BACKEND_KEY) {
+			} else if (backendKeys.indexOf(req.get('FT-Next-Backend-Key')) !== -1) {
 				res.set('FT-Backend-Authentication', true);
 				next();
 			} else {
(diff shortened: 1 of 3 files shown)


matthew-andrews:
Ahh nice, much neater/clearer than blue/green +1

keithamus:
I deliberated over the name. Ultimately went with `_OLD`/`_OLDEST` because they better indicate the age of the keys which will eventually roll out. Still not 100% on the names though.

matthew-andrews:
cc @wheresrhys any thoughts?

wheresrhys:
What if e.g. the oldest key gets leaked, then the value it's replaced with will actually be the newest key, won't it? Is there a reason for not just using FT_NEXT_BACKEND_KEY_1 etc.?

No strong opinion from me though

keithamus:
> What if e.g. the oldest key gets leaked, then the value it's replaced with will actually be the newest key, won't it?  

The point here is that, say we have `FT_NEXT_BACKEND_KEY = 123` but we want to roll the keys, so now `FT_NEXT_BACKEND_KEY = 456` and `FT_NEXT_BACKEND_KEY_OLD = 123`, eventually the oldest keys fall of (once we have rolled 4 times). 

If a key is leaked, we can revoke it across all properties, similar to how we'd revoke keys now. The difference this adds is a grace period for rolling known secure keys for all services.

>  Is there a reason for not just using FT_NEXT_BACKEND_KEY_1 etc.?

Personally I didn't go with the name `_1`/`_2` because it doesn't imply ageing of the keys.

matthew-andrews:
I think it's probably OK to just have two (current and old) but doesn't harm to have the option of a third

main.js
```diff
@@ -18,6 +18,16 @@ const serviceMetrics = require('./src/service-metrics');
...
+const backendKeys = [];
+if (process.env.FT_NEXT_BACKEND_KEY) {
+	backendKeys.push(process.env.FT_NEXT_BACKEND_KEY);
+}
+if (process.env.FT_NEXT_BACKEND_KEY_OLD) {
+	backendKeys.push(process.env.FT_NEXT_BACKEND_KEY_OLD);
+}
+if (process.env.FT_NEXT_BACKEND_KEY_OLDEST) {
+	backendKeys.push(process.env.FT_NEXT_BACKEND_KEY_OLDEST);
+}
```
phamann:
```js
Object.keys(process.env)
    .filter(key => /FT_NEXT_BACKEND_KEY/.test(key))
    .forEach(key =>  backendKeys.push(process.env[key]));
```

Sorry bit late to the game, but this is a lot cleaner and allows for future addition of keys with same prefix 😄 :

main.js
```diff
@@ -90,7 +100,7 @@ module.exports = function(options) {
 				// allow healthchecks etc. through
 				req.path.indexOf('/__') === 0) {
 				next();
-			} else if (req.get('FT-Next-Backend-Key') === process.env.FT_NEXT_BACKEND_KEY) {
+			} else if (backendKeys.indexOf(req.get('FT-Next-Backend-Key')) !== -1) {
```
phamann: This could also be a `.some()` :wink:

PR #345, opened 2016-06-09
timing out requests

wheresrhys:
In line with what Niklas told us about 5s limit on request handling

@phamann

### DIFF
--- a/main.js
+++ b/main.js
@@ -188,6 +188,19 @@ module.exports = function(options) {
 
 	let flagsPromise = Promise.resolve();
 
+	app.use((req, res) => {
+		setTimeout(() => {
+			nLogger.error('RESPONSE_TIMED_OUT', {
+				url: req.url
+			});
+			try {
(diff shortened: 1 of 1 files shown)


main.js
```diff
@@ -188,6 +188,19 @@ module.exports = function(options) {
...
 
+	app.use((req, res) => {
+		setTimeout(() => {
+			nLogger.error('RESPONSE_TIMED_OUT', {
+				url: req.url
+			});
+			try {
+				res.sendStatus(504)
+			} catch (e) {
+				// avoids sending a 'headers already sent' error to sentry
```
matthew-andrews: hmm, i think this isn't where that error would be sent from

matthew-andrews: isn't just*

matthew-andrews: also can this be an optional feature for now while we test it in a non-platinum app?

wheresrhys: sure

PR #346, opened 2016-06-13
0 is falsey in JavaScript

matthew-andrews:
so make the max-age check a bit cleverer so… that we don't forget to set it when it's 0

oh javascript.

### DIFF
--- a/src/middleware/cache.js
+++ b/src/middleware/cache.js
@@ -30,7 +30,9 @@ const presets = {
 
 function headerObjToString(obj) {
 	const parts = [];
-	if (obj['max-age']) {
+
+	// HACK: Use someething smarter than just an if (obj['max-age']) { because in JavaScrypt 0 is *falsey*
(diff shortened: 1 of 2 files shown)


src/middleware/cache.js
```diff
@@ -30,7 +30,9 @@ const presets = {
 
 function headerObjToString(obj) {
 	const parts = [];
-	if (obj['max-age']) {
+
+	// HACK: Use someething smarter than just an if (obj['max-age']) { because in JavaScrypt 0 is *falsey*
+	if (obj.hasOwnProperty('max-age')) {
```
wheresrhys: max-age in obj?

wheresrhys:
:+1:

src/middleware/cache.js
matthew-andrews: i knew there'd be some hipster way of doing this

matthew-andrews:
and again with more hipster^^

PR #347, opened 2016-06-15
decorate responses with nUiVersion

wheresrhys:
@i-like-robots @ironsidevsquincy


wheresrhys:
to support this https://github.com/Financial-Times/n-ui/pull/110

PR #353, opened 2016-06-20
Use .gitignore to set built asset expectations

wheresrhys:
@debugwand FYI I'm doing this for totally unrelated reasons to what we were discussing, but by the end of this PR it should be a little eeasier to decouple hashed assets from n-express


wheresrhys:
@ironsidevsquincy @matthew-andrews Here's the meat of changing how we check if expected assets have been built - Looks for files in .gitignore and in /public directory ending in .css or .js, and verifies both lists are the same

wheresrhys:
beginnings of a tidy up too

PR #367, opened 2016-07-12
New Navigation Model

aintgoin2goa:
Removing the n-navigation middleware and replacing it with a navigationModel that gets it's data from the new API


main.js
```diff
@@ -141,13 +143,14 @@ module.exports = function(options) {
...
 			viewsDirectory: options.viewsDirectory
-		});
+		}));
 	}
 
 	// add statutory metadata to construct the page
 	if (options.withNavigation) {
-		flagsPromise.then(navigation.init);
-		app.use(navigation.middleware);
+		let navigation = new NavigationModel({withNavigationHierarchy:options.withNavigationHierarchy});
```
i-like-robots: `const navigation = new NavigationModel(options)`?

src/navigation/hierarchyMixin.js
```diff
@@ -0,0 +1,61 @@
...
+
+	return {item:null, parent:null};
+}
+
+module.exports = class HierarchyMixin {
+
+	constructor(){
+		this.poller = new Poller({
+			url: 'http://next-navigation.ft.com/v1/taxonomy',
+			refreshInterval: ms('15m')
```
i-like-robots: Cool module but is it worth it over `1000 * 60 * 15`?

main.js
```diff
@@ -29,6 +29,7 @@ module.exports = function(options) {
 		withFlags: false,
 		withHandlebars: false,
 		withNavigation: false,
+		withNavigationHierarchy: false,
```
i-like-robots: taxonomy?

src/navigation/decorate.js
```diff
@@ -0,0 +1,61 @@
...
+
+	return false;
+}
+
+function decorateDrawerItems(items, currentUrl){
+	return items.some(item => decorateDrawerItem(item, currentUrl));
+}
+
+function decorateDrawer(items, currentUrl){
+	items.some(section => decorateDrawerItems(section, currentUrl));
```
i-like-robots: why `some`?

src/navigation/decorate.js
```diff
@@ -0,0 +1,61 @@
+"use strict";
+
+function isCurrentLink(item, currentUrl){
+	return (item.href && item.href === currentUrl) || (item.id && currentUrl.includes(item.id));
+}
+
+function decorateDrawerItem(item, currentUrl){
```
i-like-robots: I don't these methods are specific to the drawer - they should be able to traverse any link structure

src/navigation/decorate.js
```diff
@@ -0,0 +1,61 @@
...
+		item.selected = true;
+		return true;
+	}
+
+	return false;
+}
+
+function decorateItemArray(items, currentUrl){
+	items.some(item => decorateItem(item, currentUrl));
+}
```
i-like-robots: These do the same?

src/navigation/hierarchyMixin.js
aintgoin2goa: I wouldn't use it client-side but on the server why now?  It brings in no other dependencies and makes the code a bit easier to understand

main.js
aintgoin2goa: Yeah, spent some time wondering about this (and googled a definition of taxonomy), hierarchy seems to make more sense here?

src/navigation/decorate.js
aintgoin2goa: So that, once we find the current link we can stop looking

src/navigation/decorate.js
aintgoin2goa: Yeah, the drawer has a different structure to the others (nested arrays, headings and `item.item`) was easier to just split it out

src/navigation/decorate.js
aintgoin2goa: Sorry?

src/navigation/hierarchyMixin.js
i-like-robots: just cold feet over left-pad-gate

PR #379, opened 2016-08-03
Fix switch edition mobile nav bug

debugwand:
Fix the bug where switching editions meant that the mobile nav no longer got rendered by checking the pathname of the current url rather than the entire url


matthew-andrews:
+1

PR #382, opened 2016-08-08
Set the ab test state as a local

adgad:
@tom-parker @GlynnPhillips @wheresrhys (to be put in a data attribute in n-ui)


tom-parker:
Looks good to me

PR #399, opened 2016-09-07
Allow assets/handlebars to be used without flags

rowanmanning:
If you create an app with the following config, then the asset middleware will error because `res.locals.flags` is not present:

```js
const app = express({
	withHandlebars: true
});
```

This PR fixes it by setting res.locals.flags if not present.


phamann:
:+1: LGTM

wheresrhys:
:+1:

I really need to pick this middleware apart at some point. It's become a bit of a monster

PR #401, opened 2016-09-13
Allow apps to disable preloading assets

rowanmanning:
For the Origami services I'm working on, we use the Build Service for assets. The asset preloading causes a stream of 404 errors in the console and canes the server with a lot of requests.

This PR adds a `withoutAssetPreloading` option. Is this something you'd be open to?

### DIFF
--- a/main.js
+++ b/main.js
@@ -43,6 +43,7 @@ module.exports = function(options) {
 		withNavigationHierarchy: false,
 		withAnonMiddleware: false,
 		withBackendAuthentication: false,
+		withAssets: options.withHandlebars || false, // TODO always default to false for next major version
 		hasHeadCss: false,
 		hasNUiBundle: false,
 		healthChecks: []
@@ -172,7 +173,9 @@ module.exports = function(options) {
 		}));
 
 		// Decorate responses with data about which assets the page needs
-		app.use(assetsMiddleware(options, directory));
+		if (options.withAssets) {
+			app.use(assetsMiddleware(options, directory));
+		}
 
 		// Handle the akamai -> fastly -> akamai etc. circular redirect bug
 		app.use(function (req, res, next) {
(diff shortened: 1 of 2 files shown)


wheresrhys:
From your description, is the problem a deeper one, that you don't want n-express to do any work at all defining css & js assets? I'd prefer a flag to do this because
a) means less work will be done by it in your app
b) leaves the door open for you to use `res.preload()` (or whatever the method's called) in your app, and pass it some build service urls :smile:

main.js
```diff
@@ -43,6 +43,7 @@ module.exports = function(options) {
 		withNavigationHierarchy: false,
 		withAnonMiddleware: false,
 		withBackendAuthentication: false,
+		withoutAssetPreloading: false,
```
matthew-andrews: normally we use `with` prefixes……  defaulted to `false` …… althoguh that would be a breaking change i think having asset preloading on by default was probably a bad move in the first place.

n-express is _supposed_ to default to doing as little as possible……

wheresrhys: good point. A lot of asset delivering functionality is now bundled inside withHandlebars.

Don't _really_ want to create a major release for this though. We could however have some logic which turns it off if we have `withAssetUrlGeneration: false`

matthew-andrews: a `withAssetUrlGeneration` option that deafults to true with a `// TODO` to make it default to `false` in the next major version sounds good 👍

rowanmanning:
Yeah I guess it is, and I would rather be able to disable all of the asset stuff yeah. I can dig through and see if I can disable assets, if you have a spare mo to point me in the right direction that'd be great :)

rowanmanning:
OK @wheresrhys, @matthew-andrews: I've updated to remove the assets middleware entirely if the `withAssets` option is `false`.

The `withAssets` option defaults to `false` unless `withHandlebars` is set, so this shouldn't be a breaking change.

wheresrhys:
:+!:

PR #403, opened 2016-09-15
Navigation fallback

aintgoin2goa:
Will allow apps to start when navigation API is down.

If the api is down the app will try to go direct to the s3 bucket.  This will mean all non-dynamic navigation will be displayed. Currently, that means everything but the meganav.

If S3 is also down then it will fallback to some hardcoded data and will display just top level section links - there will be no footer, no account links and no child links in the drawer.

We probably need a discussion to decide how many links to hard code here as there will be effort involved in keeping them up to date.

@commuterjoy 
@i-like-robots


commuterjoy:
double protection :)

> how many links to hard code 

Just the top level ones (Home -> Life & Arts) in the local default & s3 are fine with me, anything else I fear will get out of sync.

+1

PR #406, opened 2016-09-16
make it possible to simulate healthcheck failures

ifyio:
cc/ @commuterjoy


commuterjoy:
I think so :)

Test it out on a smaller app first from this branch?

Could we have the same thing for gtg too? Later.

PR #408, opened 2016-09-21
Nui option

lc512k:
@wheresrhys I've added the hasNUiBundle option back in, but defaulting to true for apps like beacon that might still want to turn it off

### DIFF
--- a/main.js
+++ b/main.js
@@ -48,8 +48,8 @@ module.exports = function(options) {
 		withNavigationHierarchy: false,
 		withAnonMiddleware: false,
 		withBackendAuthentication: false,
-		// TODO always default to false for next major version
 		hasNUiBundle: true,
+		// TODO always default to false for next major version
 		withAssets: options.withHandlebars || false,
 		hasHeadCss: false,
 		healthChecks: []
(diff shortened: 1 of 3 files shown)


main.js
```diff
@@ -49,6 +49,7 @@ module.exports = function(options) {
 		withAnonMiddleware: false,
 		withBackendAuthentication: false,
 		// TODO always default to false for next major version
```
wheresrhys: could you move this comment down a line as it should be with the withAssets option

src/middleware/assets.js
```diff
@@ -103,12 +103,17 @@ Or \`rm -rf bower_components/n-ui && bower install n-ui\` if you're no longer wo
...
-			res.locals.nUiConfig = nUiConfig;
-			res.locals.javascriptBundles.push(`\
-${nUiUrlRoot}\
-es5-${res.locals.flags.polyfillSymbol ? 'polyfill-io' : 'core-js'}\
-${(res.locals.flags.nUiBundleUnminified || nUiIsLinked ) ? '' : '.min'}.js`);
-			res.locals.javascriptBundles.push(hashedAssets.get('main-without-n-ui.js'));
+			if (options.hasNUiBundle) {
+				res.locals.nUiConfig = nUiConfig;
+				res.locals.javascriptBundles.push(`\
+				${nUiUrlRoot}\
```
wheresrhys: having no indentation here is important or it generates a url with lots of whitespace in it. If that's confusing then it might be better to construct the string from an array joined with `''`?

PR #413, opened 2016-09-26
System code mandatory

commuterjoy:
As discussed in the huddle. I will merge this a few weeks after default day.


commuterjoy:
merging on 18th

PR #414, opened 2016-09-27
Add opt out link for footer

debugwand:
https://github.com/Financial-Times/n-ui/pull/494

### DIFF
--- a/src/navigation/decorate.js
+++ b/src/navigation/decorate.js
@@ -36,6 +36,6 @@ function decorateItems(items, currentUrl){
 	}
 }
 
-module.exports = function decorate(items, listName, currentUrl) {
+module.exports = function decorate(items, currentUrl) {
 	return decorateItems(items, currentUrl);
 };
--- a/src/navigation/decorate.js
+++ b/src/navigation/decorate.js
@@ -37,10 +37,5 @@ function decorateItems(items, currentUrl){
 }
 
 module.exports = function decorate(items, listName, currentUrl) {
-	switch (listName) {
-		case "footer":
-			return;
-		default:
-			return decorateItems(items, currentUrl);
-	}
+	return decorateItems(items, currentUrl);
 };
--- a/src/navigation/optOutFooter.js
+++ b/src/navigation/optOutFooter.js
@@ -1,0 +1,16 @@
+"use strict";
+
+// this provides an opt-out link in the footer for the compulsory opt-in phase of Next
+module.exports = function (items, currentUrl) {
+	const optOutLink = {
+		name: 'Old FT.com',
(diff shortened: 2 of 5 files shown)


src/navigation/decorate.js
```diff
@@ -37,10 +37,5 @@ function decorateItems(items, currentUrl){
 }
 
 module.exports = function decorate(items, listName, currentUrl) {
```
keirog: `listName` not needed now

keirog: oh, I guess that was to keep the signature the same?

src/navigation/optOutFooter.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = function (items, currentUrl) {
+	const optOutLink = {
+		name: 'Old FT.com',
+		href: `/opt-out-confirm?location=${encodeURIComponent(currentUrl)}`
+	};
+	const supportBlock = items.find(i => {
+		return i.title === 'Support';
+	});
+	if (supportBlock) {
+		supportBlock.optOutLink = optOutLink;
```
keirog: How is `supportBlock` then used?

src/navigation/optOutFooter.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+// this provides an opt-out link in the footer for the compulsory opt-in phase of Next
+module.exports = function (items, currentUrl) {
+	const optOutLink = {
+		name: 'Old FT.com',
```
keirog: I'd favour adding a `trackableName` (or similar) for use in the `data-trackable` attributes, rather than reusing `name`. That way if the copy changes, we don't have to change our beacon queries to include the amended names.

src/navigation/optOutFooter.js
debugwand: you can see in the n-ui footer pr, if this item is present it outputs it to the top of the list of links if certain conditions are true

src/navigation/decorate.js
debugwand: I can remove that I think

src/navigation/optOutFooter.js
keirog: Sorry I mean I don't understand how that's exported

debugwand: ah - this module receives `items` as an argument (items are 'support', 'legal and services', 'tools' etc), looks for the one title support, then adds the optout link to that block, before returning the amended `items`. Footer has items which have their own items.

keirog: Oh I see, I didn't grok properly that `supportBlock` was the found object. Ta!

src/navigation/optOutFooter.js
debugwand: done

PR #425, opened 2016-11-07
Add node_modules to handlebars partials locations

andygnewman:
cc: @i-like-robots 

Enable the finding of partials in node_modules

### DIFF
--- a/src/handlebars/index.js
+++ b/src/handlebars/index.js
@@ -14,7 +14,7 @@ module.exports = function (conf) {
 	return handlebars(app, {
 		partialsDir: [
 			directory + (options.viewsDirectory || '/views') + '/partials',
-			directory + ('/node_modules/@financial-times')
+			directory + ('/node_modules')
 		],
 		defaultLayout: false,
 		// The most common use case, n-ui/layout is not bundled with this package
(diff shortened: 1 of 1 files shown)


src/handlebars/index.js
```diff
@@ -13,7 +13,8 @@ module.exports = function (conf) {
 
 	return handlebars(app, {
 		partialsDir: [
-			directory + (options.viewsDirectory || '/views') + '/partials'
+			directory + (options.viewsDirectory || '/views') + '/partials',
+			directory + ('/node_modules/@financial-times')
```
i-like-robots: I wonder if this should look in a subfolder to reduce cost at startup? Maybe `/views`? Might be good to have a convention, what do you think?

andygnewman: If we made it more specific we'd have to do a glob to find the specific ones and change the referencing to be less specific.

With the specific reference in the template to the partial I don't think that this should be costly at start up as should only be invoked when looking for a specific partial.

src/handlebars/index.js
```diff
@@ -13,7 +13,8 @@ module.exports = function (conf) {
 
 	return handlebars(app, {
 		partialsDir: [
-			directory + (options.viewsDirectory || '/views') + '/partials'
+			directory + (options.viewsDirectory || '/views') + '/partials',
+			directory + ('/node_modules')
```
debugwand: again, limit to \@financial-times

andygnewman:
Now scoped the node partials to `/@financial-times`

PR #429, opened 2016-11-27
Adds a new option `partialsDirectory`

GlynnPhillips:
Will allow an additional path to load partials from.

I have added this as it would be required for pa11y testing of n-ui but
I can also imagine other uses when our normal convention might not
always work.

@wheresrhys @andygnewman Sorry picking on you as you have contributed to this file

/cc @lc512k


andygnewman:
Am generally a fan of sticking to conventions, so may be worth in the README highlighting the conventions.

Also is a little strange in setting up the conventions we do the defaults for views/partials and node_modules/@financial-times in n-express and the bower_components in n-handlebars - feels like they should be in one place.

But no objection if the README is clear on the conventions.

GlynnPhillips:
@andygnewman I agree we should be clear on this and I did find some weirdness like the point you make about bower components when looking into this feature. 

I think documenting the conventions though should be down outside of this commit especially as I would not be familiar with all of them so don't feel like I would be able to document them correctly. Seems like the README has been left behind and we would actually need a big clean up

PR #438, opened 2016-12-14
Adds a basic WebPage schema to all requests

GlynnPhillips:
This will allow any app that wants to display a basic WebPage schema for
FT pages to do so.

The the templating is handled by `wrapper.html` in `n-ui` and each app
will just need add this data to a `jsonLd` key in the json passed to
render.

### DIFF
--- a/main.js
+++ b/main.js
@@ -9,6 +9,7 @@ const express = require('express');
 const http = require('http');
 const https = require('https');
 const denodeify = require('denodeify');
+const nextJsonLd = require('next-json-ld');
 
 const flags = require('next-feature-flags-client');
 const backendAuthentication = require('./src/middleware/backend-authentication');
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -144,6 +145,11 @@ module.exports = function(options) {
 		next();
 	});
 
+	app.use(function(req, res, next) {
+		res.locals.jsonLd = [nextJsonLd.webSite()];
+		next();
+	});
```
ironsidevsquincy: do we want to make this an 'opt in' option - https://github.com/Financial-Times/n-express/blob/0ce7743c8a0d768a717b090a2fd59f04f855fc06/main.js#L53-L64

wheresrhys:
Have you checked that it doesn't bleed from one request to another. Also woudl be good to add to the README. Otherwise :+1:

main.js
wheresrhys: good point quincy. If it's on by default then it adds extra processing to do for high volume apis such as preflight, which isn't very desirable.

I'm thinking of abstracting all the user-facing app stuff into something like n-html-app early next year, so at that point could decide to make it the default

ironsidevsquincy: that sounds good

@GlynnPhillips, was thinking about APIs, things that don't output any using facing html but still use n-express

PR #451, opened 2017-01-17
tests for backend authentication configured entirely by presence of keys

wheresrhys:
This will be a major release


sjparkinson:
So we have this and also #455.

Seeing as `withBackendAuthentication` defaults to false, perhaps a default healthcheck would be better to verify everything is OK when it is enabled. Rather than fail open here?

wheresrhys:
not sure I follow

PR #454, opened 2017-02-09
remove `time` dependency as it is not architecture independant

bjfletcher:
we can't reuse slug built locally on Heroku because of this - you get the `elf` error


sjparkinson:
I'd love to see this go out, would allow next-es-interface to use node 7 I believe.

doramatadora:
Closing as we removed `time`

PR #455, opened 2017-02-23
IP address whitelist fallback if backend keys don't work

bjfletcher:
Plan:

1. merge and release https://github.com/Financial-Times/ft.com-cdn/pull/328 first
2. merge and release this
3. get stream app (bronze) to use this new version
4. if all ok, then get frontpage & article apps to use this new version


src/middleware/backend-authentication.js
```diff
@@ -19,17 +23,19 @@ module.exports = appName => {
...
 			res.set('FT-Backend-Authentication', true);
 			next();
 		} else {
 			res.set('FT-Backend-Authentication', false);
 			if (process.env.NODE_ENV === 'production') {
 				// NOTE - setting the status text is very important as it's used by the CDN
-				// to trigger stale-if-error if we mess up the key synchronisation again
-				res.status(401).send('Invalid Backend Key');
+				// to trigger stale-if-error
+				res.status(401).send('Invalid Backend Authentication');
```
wheresrhys: the value of this is important - it causes 401 to be converted to 503 in fastly, so (when we get  cching right) should mean stale-if-error kicks in.

Interesting question - what should we do when we move over to IP whitelist?

src/lib/ip-whitelist-backup.json
```diff
@@ -0,0 +1,18 @@
+{
+	"addresses": [
+		"23.235.32.0/20",
```
wheresrhys: is it wise to store this list in here. Is there potential for the values to be spoofed? config-vars feels like a more natural home, and also allows tying it in to sam's auto key rolling work

src/lib/ip-whitelist.js
```diff
@@ -0,0 +1,42 @@
...
+const ip = require('ip');
+
+const backupWhitelist = require('./ip-whitelist-backup.json').addresses;
+
+const whitelistUrl = 'https://api.fastly.com/public-ip-list';
+
+function IpWhitelist () {
+	this.fetchedWhitelist = null;
+	this.poll();
+	setInterval(() => this.poll(), 10000); // every 10 seconds
```
wheresrhys: fyi, we have the ft-poller component, which you don't have to use, but it saves writing all this fiddly polling stuff

src/lib/ip-whitelist.js
```diff
@@ -0,0 +1,42 @@
...
+				logger.error({ event: 'IP_WHITELIST_UNRECOGNISED', response: JSON.stringify(resp) });
+			}
+		})
+		.catch(err => logger.error({ event: 'IP_WHITELIST_FETCH_FAIL' }, err));
+}
+
+IpWhitelist.prototype.validate = function (ipAddress) {
+	const ranges = this.fetchedWhitelist || backupWhitelist;
+	let i;
+	for (i = 0; i < ranges.length; i++) {
```
wheresrhys: this will happen on every request, so perhaps a rare case where we should micro-optimise. Is string-matching against a `,` concatenated list quicker? or a map of {ipAddress1: true, ipAddress2: true} etc

src/lib/ip-whitelist.js
```diff
@@ -0,0 +1,42 @@
...
+			if (Array.isArray(resp.addresses) && resp.addresses.length > 0) {
+				if (JSON.stringify(this.fetchedWhitelist) !== JSON.stringify(resp.addresses)) {
+					logger.info({ event: 'IP_WHITELIST_UPDATE', oldSize: Array.isArray(this.fetchedWhitelist) ? this.fetchedWhitelist.length : 0, newSize: resp.addresses.length });
+					this.fetchedWhitelist = resp.addresses;
+				}
+			} else {
+				logger.error({ event: 'IP_WHITELIST_UNRECOGNISED', response: JSON.stringify(resp) });
+			}
+		})
+		.catch(err => logger.error({ event: 'IP_WHITELIST_FETCH_FAIL' }, err));
```
wheresrhys: can we send metrics for both of these too, as metrics are likely to underpin any healthchecks

src/middleware/backend-authentication.js
bjfletcher: Have you seen this? https://github.com/Financial-Times/ft.com-cdn/pull/328

wheresrhys: OK good. But I think that PR should accept either value for a while as you won't be able to roll out to all the apps + the CDN simultaneously

src/lib/ip-whitelist.js
bjfletcher: Good call, I'll add metrics in.

src/lib/ip-whitelist.js
bjfletcher: aye, I didn't particularly like the look of `ft-poller` - it doesn't have the main logic in one place and also it doesn't do any logging... it's smelly :)

src/lib/ip-whitelist.js
bjfletcher: I'm not sure how we can match subnets? For example, `23.235.32.0/20`. A list of IP addresses, great idea... but a list of subnets? BTW, this runs only if backend keys don't work.

src/middleware/backend-authentication.js
bjfletcher: Great spot Rhys. I've just updated the Fastly PR to accept both

src/lib/ip-whitelist-backup.json
aintgoin2goa: There's an API for getting this https://docs.fastly.com/api/tools#public_ip_list

src/lib/ip-whitelist.js
aintgoin2goa: There are modules for doing that on npm.  FWIW  I tried to implement this functionality a couple of years ago and it was really hard.  I have faith you will succeed though! 🍀

PR #457, opened 2017-03-01
Remove time

sjparkinson:
Relates to #454, might be overkill though.

### DIFF
--- a/src/lib/check-failing.js
+++ b/src/lib/check-failing.js
@@ -1,5 +1,5 @@
-const time = require('time');
 const isWithinRange = require('date-fns/is_within_range');
+const fs = require('fs');
 
 const the = {
 	currentDate: new Date(),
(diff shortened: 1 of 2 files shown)


sjparkinson:
The only place that'd use this information anyway is [next-encryption-api](https://github.com/Financial-Times/next-encryption-api), [and it doesn't seem to](https://github.com/Financial-Times/next-encryption-api/search?utf8=%E2%9C%93&q=timezone).

src/lib/check-failing.js
```diff
@@ -7,8 +7,23 @@ const the = {
 	failures: []
 };
 
-module.exports.init = function () {
+let getTimezone = () => {
```
bjfletcher: go for immutability where possible... `const` instead of `let`

src/lib/check-failing.js
```diff
@@ -7,8 +7,23 @@ const the = {
...
+	if (process.env.TZ) {
+		return process.env.TZ;
+	}
+
+	if (fs.existsSync('/etc/timezone')) {
+		return fs.readFileSync('/etc/timezone');
```
bjfletcher: you reckon this file can't be blank?

src/lib/check-failing.js
```diff
@@ -7,8 +7,23 @@ const the = {
...
+	}
+
+	if (fs.lstatSync('/etc/localtime').isSymbolicLink()) {
+		return fs.readlinkSync('/etc/localtime').replace('/usr/share/zoneinfo/', '');
```
bjfletcher: ditto, worth not returning if it's blank

src/lib/check-failing.js
```diff
@@ -7,8 +7,23 @@ const the = {
...
+	}
+
+	console.error('Error occurred when evaluating the system timezone'); // eslint-disable-line
```
bjfletcher: I'd replace with:
`logger.error({ event: 'EXPRESS_GET_TIMEZONE_FAIL' });`
then return `new Date().getTimezone()` (or whatever is the fix in my PR)

PR #464, opened 2017-05-10
Add default error rate healthcheck

TheoLeanse:


src/lib/health-checks.js
```diff
@@ -33,9 +40,9 @@ module.exports = (app, options, meta) => {
 			schemaVersion: 1,
 			name: options.healthChecksAppName || defaultAppName,
 			systemCode: options.systemCode,
-			description:meta. description,
+			description: meta.description,
```
TheoLeanse: not sure why this didn't error previously?

test/fixtures/app/main.js
```diff
@@ -5,7 +5,8 @@ const express = require('../../..');
 const app = module.exports = express({
 	directory: __dirname,
 	withFlags: !process.env.DISABLE_FLAGS,
-	systemCode: 'test-app'
+	systemCode: 'test-app',
+	skipDefaultErrorRateCheck: true
```
TheoLeanse: should perhaps have two app fixture, one with default check and one without

Makefile
```diff
@@ -13,12 +13,10 @@ else
 endif
 
 coverage-report: ## coverage-report: Run the unit tests with code coverage enabled.
-	unset FT_NEXT_BACKEND_KEY && istanbul cover node_modules/.bin/_mocha --report=$(if $(CIRCLECI),lcovonly,lcov) 'test/**/*.test.js'
+	unset FT_NEXT_BACKEND_KEY && FT_GRAPHITE_KEY=foobar istanbul cover node_modules/.bin/_mocha --report=$(if $(CIRCLECI),lcovonly,lcov) 'test/**/*.test.js'
```
remybach: Just curious, but why is this necessary?

test/fixtures/app/main.js
remybach: I'm inclined to agree with that.

Makefile
TheoLeanse: I've introduced the `n-health` module which blows up if that's not present:

screen shot 2017-05-11 at 09 48 13

remybach: Fair enough 👍

test/app/health.test.js
```diff
@@ -1,25 +1,52 @@
...
+
+		it('should not 500 /__health.1', function (done) {
+			request(defaultApp)
+				.get('/__health.1')
+				.expect(200, done);
+		});
+
+		it('should 500 /__health.2', function (done) {
+			// will 500 because it has a level 2 error rate healthcheck by
+			// default which has not been run at this point and thus is not OK
```
TheoLeanse: this is the case, but I'm not sure if this is OK...

Makefile
adgad: Do all apps have FT_GRAPHITE_KEY set? I assume they'd also blow up if not...

src/lib/error-rate-check.js
```diff
@@ -0,0 +1,23 @@
+const nHealth = require('n-health');
+
+module.exports = (appName) => (
+	nHealth.runCheck({
+		type: 'graphiteThreshold',
+		metric: `
+			divideSeries(
+				sumSeries(
+					next.heroku.${appName}.web_*_${process.env.REGION || '*'}.express.default_route_GET.res.status.500.count
+				),
```
adgad: would the `web_*_*` break if there's no region?

src/lib/error-rate-check.js
```diff
@@ -0,0 +1,23 @@
...
+				sumSeries(
+					next.heroku.${appName}.web_*_${process.env.REGION || '*'}.express.default_route_GET.res.status.*.count
+				)
+			)
+		`.replace(/\t|\n/g, ''),
+		threshold: 0.02,
```
adgad: Wonder if this is quite a low threshold for a default? What's the rate for something like myFT or bronzier apps?

src/lib/error-rate-check.js
```diff
@@ -0,0 +1,23 @@
+const nHealth = require('n-health');
+
+module.exports = (appName) => (
+	nHealth.runCheck({
+		type: 'graphiteThreshold',
+		metric: `
+			divideSeries(
+				sumSeries(
+					next.heroku.${appName}.web_*_${process.env.REGION || '*'}.express.default_route_GET.res.status.500.count
```
ironsidevsquincy: maybe `*_GET`, to capture all routes

src/lib/error-rate-check.js
```diff
@@ -0,0 +1,23 @@
...
+				),
+				sumSeries(
```
wheresrhys: maybe a moving average over 5 or 10 minutes

src/lib/health-checks.js
```diff
@@ -1,13 +1,20 @@
...
+const errorRateCheck = require('./error-rate-check');
 
 module.exports = (app, options, meta) => {
-	const healthChecks = options.healthChecks;
 	const defaultAppName = `Next FT.com ${meta.name} in ${process.env.REGION || 'unknown region'}`;
 
+	const defaultChecks = [];
+	if (!options.skipDefaultErrorRateCheck) {
+		defaultChecks.push(errorRateCheck(meta.name));
+	}
```
ironsidevsquincy: maybe we shouldn't give the option to turn this off, just the option to set the threshold...

src/lib/error-rate-check.js
```diff
@@ -0,0 +1,23 @@
...
+					next.heroku.${appName}.web_*_${process.env.REGION || '*'}.express.default_route_GET.res.status.*.count
+				)
+			)
+		`.replace(/\t|\n/g, ''),
+		threshold: 0.02,
+		name: `500 rate for ${appName} is acceptable`,
+		severity: 2,
```
wheresrhys: I would default this to 3 (or 1 - I can't remember which is the highest level), but possibly allow configuring it as an n-express option - some apps (particularly experimental ones) may error a lot

src/lib/error-rate-check.js
wheresrhys: good to make this configurable too

src/lib/error-rate-check.js
```diff
@@ -0,0 +1,23 @@
...
+		businessImpact: `Error rate for the ${appName} app is higher than the acceptable threshold of 2%.`,
+		technicalSummary: `The proportion of 500 responses for the ${appName} application requests across all heroku dynos vs all responses is higher than a threshold of 0.02`,
+		panicGuide: 'Consult application logs in splunk and run the application locally to identify errors'
```
wheresrhys: add link to sentry too

src/lib/error-rate-check.js
TheoLeanse: myFT uses a `graphiteSpike` check (threshold 5) rather than a `graphiteThreshold` check

TheoLeanse: not sure what a sensible default threshold might be... 0.5?

src/lib/error-rate-check.js
TheoLeanse: more like this: https://github.com/Financial-Times/next-myft-page/blob/master/server/health-checks/50x.js ?

Makefile
TheoLeanse: not in dev - `n-express` only has a config-vars entry for ci

src/lib/error-rate-check.js
TheoLeanse: also not totally clear on the relative merits of `graphiteThreshold` and `graphiteSpike`

src/lib/error-rate-check.js
```diff
@@ -0,0 +1,29 @@
+const nHealth = require('n-health');
+
+const DEFAULT_THRESHOLD = 0.04;
+const DEFAULT_SEVERITY = 1;
```
wheresrhys: on reflection, at least while trialling this, we should probably set it to low severity

TheoLeanse: done

PR #479, opened 2017-10-25
Security headers

sjparkinson:
See https://securityheaders.io/?q=https%3A%2F%2Fwww.ft.com&hide=on.

Things to confirm:

- [ ] ~Referrer-Policy `strict-origin-when-cross-origin` is 🆗  with systems like spoor?~
- [x] X-Frame-Options `SAMEORIGIN` is 🆗 , is www.ft.com loaded in iframes anywhere?

### DIFF
--- a/main.js
+++ b/main.js
@@ -16,9 +16,10 @@ const healthChecks = require('./src/lib/health-checks');
 const instrumentListen = require('./src/lib/instrument-listen');
 const guessAppDetails = require('./src/lib/guess-app-details');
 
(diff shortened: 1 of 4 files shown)


sjparkinson:
First I'm going to test these out on next-stream-page.

sjparkinson:
Not going to include `Referrer-Policy` as it'll have an impact on tracking between https://www.ft.com and https://markets.ft.com for example.

PR #484, opened 2017-11-28
Remove no-store directive

### LINKED ISSUE #483: Enable 304 responses
I believe the use of `no-store` is causing us to force full reloads of pages (e.g. https://www.ft.com/) when we could be responding with 304 responses.

Is this intentional?

https://github.com/Financial-Times/n-express/blob/b604fbf3b5d23c81d48d74800fcf69986292f5df/src/middleware/cache.js#L2

sjparkinson:
Resolves #483.

Enables 304 not modified responses.

### DIFF
--- a/src/middleware/cache.js
+++ b/src/middleware/cache.js
@@ -1,5 +1,5 @@
 module.exports = (req, res, next) => {
-	res.FT_NO_CACHE = 'max-age=0, no-cache, no-store, must-revalidate';
+	res.FT_NO_CACHE = 'max-age=0, no-cache, must-revalidate';
 	res.FT_SHORT_CACHE = 'max-age=600, stale-while-revalidate=60, stale-if-error=86400';
 	res.FT_HOUR_CACHE = 'max-age=3600, stale-while-revalidate=60, stale-if-error=86400';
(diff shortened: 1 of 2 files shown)


src/middleware/cache.js
```diff
@@ -1,5 +1,5 @@
 module.exports = (req, res, next) => {
-	res.FT_NO_CACHE = 'max-age=0, no-cache, no-store, must-revalidate';
+	res.FT_NO_CACHE = 'max-age=0, no-cache, must-revalidate';
```
wheresrhys: I'm sure there's some danger here. @commuterjoy was always very opinionated about really strict non-caching

sjparkinson: Yep I'm sure there's a reason, but don't know what it is.

Personally I don't see an issue with 304 responses, but willing to be shown otherwise.

wheresrhys: I think it's due to intermediate caches not necessarily respecting vary, so we need to force intermediates not to store, and no-store is the only bulletproof way to do so

commuterjoy: I think it's safe to remove, I see other sites have simplified their no-cache directives too. IIRC the original FT no cache was based on a bit of an audit of what other personalised sites do. If you look at github, google etc. it's often just `max-age=0` these days.

PR #485, opened 2017-11-29
remove healthcheck failure simulation

wheresrhys:
@bjfletcher @sjparkinson 

I don't think anyone's used this for a very long time and it can cause weird test failures in apps' test suites.


sjparkinson:
<3

PR #487, opened 2018-01-02
Modify the place to send unrecognised services' error from Sentry to HealthCheck

fenglish:


src/lib/unregistered-services-healthCheck.js
```diff
@@ -0,0 +1,36 @@
...
+	setAppName: appName => {
+		return {
+			getStatus: () => {
+				return {
+					name: `All services for ${appName} are registered in Next-metrics`,
+					ok: lastCheckOk,
+					checkOutput: lastCheckOutput,
+					lastUpdated: lastCheckTime,
+					panicGuide: panicGuide,
+					severity: 3,
```
fenglish: Is this correct severity?

test/app/app.test.js
```diff
@@ -137,28 +137,8 @@ describe('simple app', function () {
 				});
 
 		});
-
-		//fixme - I'm not sure ow this test ever passed but it doesn't now
-		it.skip('should notify sentry of unrecognised services', function (done) {
-
```
fenglish: I delete this test do I need a test for new code?

wheresrhys: no you can delete it safely. Wasn't running anyway

src/lib/unregistered-services-healthCheck.js
wheresrhys: I can never remember if 3 is the most severe or least :joy:

If the existing error rate check is 3, then that means 3 is probably most severe. 

Your new check should be least severe, so if the error rate check is 3, make the new one a 1

PR #488, opened 2018-01-10
Remove the failing default check

sjparkinson:
Fixes:

image

Is this what we're happy with? There are still two default checks in each app.

Does this stop nudging people to add checks that apps should have?

### DIFF
--- a/src/lib/health-checks.js
+++ b/src/lib/health-checks.js
@@ -14,17 +14,7 @@ module.exports = (app, options, meta) => {
 	app.get(/\/__health(?:\.([123]))?$/, (req, res) => {
 		res.set({ 'Cache-Control': 'private, no-cache, max-age=0' });
 		const checks = healthChecks.map(check => check.getStatus());
-		if (checks.length === defaultChecks.length) {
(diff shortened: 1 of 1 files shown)


adambraimbridge:
It's worth noting: 

- These healthchecks are checked because their respective `__health` endoints have been added to the CMDB endpoints table.  Before the endpoints were added to CMDB, there were no alerts.
- The endpoints in CMDB have an `isLive:[True|False]` property, which is being ignored by this healthcheck aggregator. Not a problem, but worth being aware of.
- Non-crucial healthcheck alerts like the ones in the above screenshot are largely ignored, and it can be argued they contribute to a "cry-wolf" effect; drowning out more urgent alerts.

#### Quis custodiet ipsos custodes
When you're guarding the guards, you can't do it from the same place. You need a separate monitor. If we want to check that we have healthchecks, we can't do that from the thing that does the healthchecks. Something like http://ombudsman.in.ft.com/ is more appropriate (which could then be exposed in http://houston.in.ft.com/ — which already has a panel for healthchecks)

PR #489, opened 2018-01-16
moving anon into here from n-ui

wheresrhys:


src/middleware/anon.js
```diff
@@ -0,0 +1,41 @@
...
+		this.userIsAnonymous = true;
+	} else {
+		this.userIsLoggedIn = true;
+		this.userIsAnonymous = false;
+	}
+}
+
+function FirstClickFreeModel () {
+	this.signInLink = '/login';
+	this.subscribeNowLink = '/product-selector?segID=400872&segmentID=676c655f-9b47-27a8-97db-ab3a6a6dbc54';
```
tavvy: shoulds this be `/products?segID=400872&segmentID=676c655f-9b47-27a8-97db-ab3a6a6dbc54` ?

wheresrhys: this is copied directly from n-ui. if you're sure it's wrong I can change in this PR if you like

tavvy: https://www.ft.com/product-selector returns a `404` /shrug

PR #495, opened 2018-02-21
Setup NO_CACHE to test the no-store directive

sjparkinson:
Pairs with the `noStoreCache` header.

### DIFF
--- a/src/middleware/cache.js
+++ b/src/middleware/cache.js
@@ -1,5 +1,10 @@
 module.exports = (req, res, next) => {
-	res.FT_NO_CACHE = 'max-age=0, no-cache, must-revalidate';
+	if (res.locals.flags && res.locals.flags.noStoreCache) {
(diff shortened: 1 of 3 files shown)


src/middleware/cache.js
```diff
@@ -1,5 +1,10 @@
 module.exports = (req, res, next) => {
-	res.FT_NO_CACHE = 'max-age=0, no-cache, must-revalidate';
+	if (res.locals.flags && res.locals.flags.noStoreCache) {
+		res.FT_NO_CACHE = 'max-age=0, no-cache, no-store, must-revalidate';
+	} else {
+		res.FT_NO_CACHE = 'max-age=0, no-cache, must-revalidate';
```
sjparkinson: Going with the new way as the default, as we've fixed all the tests etc. in dependent apps.

sjparkinson:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

https://stackoverflow.com/a/11229473

PR #499, opened 2018-04-16
First pass at blocking googleweblight crawler

remybach:
"First pass" because I'm not entirely convinced this will work. Google's [official recommendation](https://support.google.com/webmasters/answer/6211428?hl=en&ref_topic=2370586) is to send a `Cache-Control: no-transform` header, which would probably have a lot of unintended side effects.

The reason for wanting to disable this is because it has no knowledge of the kind of access a user might have and therefore is hitting the barrier (which is subsequently tanking conversion metrics). Below is a graph representing the volume of page views it's hitting the site with:

[]( https://beacon.ft.com/data/query-wizard?query=page%3Aview-%3Ecount()-%3Efilter(device.userAgent~googleweblight)-%3ErelTime(previous_8_weeks)-%3Einterval(w)-%3Etidy() )

🐿 v2.8.0

### DIFF
--- a/src/assets/robots.txt
+++ b/src/assets/robots.txt
@@ -1,5 +1,8 @@
 User-agent: Twitterbot
 Disallow:
+
+User-agent: googleweblight
+Disallow:
  
 User-agent: *
 Disallow: /


adgad:
Shouldn't it be `Disallow: /` to block access? I think `Disallow: ` allows everything... https://varvy.com/robottxt.html

remybach:
Derp, you would be correct. Fix incoming...

remybach:
I've realised that this is a change that needs to happen here instead: https://github.com/Financial-Times/ft.com-cdn/blob/7782445d998d5d479f005a14f31f87dfb5c21057/src/vcl/next-robots.vcl

PR #502, opened 2018-06-08
No unregistered service reset

sjparkinson:
This check is _really_ flappy.

My thinking behind not resetting is that we don't need to. If we've fixed this check, then that'll be via a deploy of the application that reboots everything anyway.

The issue with resetting is in a given minute we won't always call a service, this is what leads this to be a flappy health check.

The alternative might be to increase the interval to 5 minutes, but there's still a chance it'd be flappy.

### DIFF
--- a/src/lib/service-metrics.js
+++ b/src/lib/service-metrics.js
@@ -7,7 +7,6 @@ module.exports = {
 		unRegisteredServicesHealthCheck.updateCheck(unregisteredServices);
 		setInterval(() => {
 			unRegisteredServicesHealthCheck.updateCheck(unregisteredServices);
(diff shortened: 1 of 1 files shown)


adambraimbridge:
I think get it. Is this about right?

 * When there _is_ a service with missing registration/s,
 * It correctly throws an error
 * But because the heathcheck resets the array of [services with missing registration/s],
 * It incorrectly thinks the error's resolved
 * Only to detect the error again because it hasn't actually been fixed.

sjparkinson:
This will mean these convert from flappy health checks to alerting health checks 😁

PR #515, opened 2018-07-25
Standardise healthcheck names and make them agnostic of state

remybach:
🐿 v2.10.1

### DIFF
--- a/src/lib/unregistered-services-healthCheck.js
+++ b/src/lib/unregistered-services-healthCheck.js
@@ -8,7 +8,7 @@ module.exports = {
 		return {
 			getStatus: () => {
 				return {
-					name: `All services for ${appName} are registered in Next-metrics`,
+					name: `Metrics: All services for ${appName} registered in next-metrics`,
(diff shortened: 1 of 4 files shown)


src/lib/unregistered-services-healthCheck.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
 		return {
 			getStatus: () => {
 				return {
-					name: `All services for ${appName} are registered in Next-metrics`,
+					name: `Metrics: All services for ${appName} registered in next-metrics`,
```
CourtneyLO: Just a small thing, should it not be: Metrics: All services for ${appName} **are** registered in next-metrics?

remybach: No, because when it fails that `are` makes it look like it's a problem that they are registered.

CourtneyLO: Oh right, that was the thing you mentioned yesterday, sorry.

PR #520, opened 2018-08-15
Add helpers to support async handlers 

maxcbc:
# Add helpers to support async handlers

## What
- Adds 2 helpers to main export for wrapping async handlers

 🐿 v2.10.2


src/lib/patch-express-async-support.js
```diff
@@ -0,0 +1,41 @@
...
+	});
+	Router.param = param;
+};
+
+function param (...args) {
+	const processedArgs = args.map(arg => {
+		if (typeof arg !== 'function') return arg;
+		return function (req, res, next, id, name) {
+			const result = arg(req, res, next, id, name);
+			if (result && result.catch) result.catch(next);
```
chee: should we use `result instanceof Promise` here, or check that `result.catch` is in fact a function before calling it or something?

i-like-robots:
Perhaps this would be nicer if it was implemented as a helper function? Folks would need to be aware of it and opt-in but that approach would avoid monkey internals.

kavanagh:
I've been using [express-async-errors](https://github.com/davidbanham/express-async-errors/blob/master/index.js), it also uses express's Layers (which seemed the right way when I looked into it last).

I think it needs to be opt in because apps already have it. And n-express is already way too opinionated which makes it a pain to use for lots of things.

But why roll our own?

maxcbc:
@kavanagh @i-like-robots I'm inclined to agree about this being rewritten as a helper.
With `express-async-errors` my only problem with that is lack of support of 'app.param' middleware, so i think its still worth rolling our own.

kavanagh:
I've never done async things in param handlers so never considered that.

maxcbc:
Have switched this to be a couple of helpers rather than a monkey patch

src/lib/async-handlers.js
```diff
@@ -0,0 +1,25 @@
...
+		return (err, req, res, next) => {
+			const result = fn(err, req, res, next);
+			if (!result.headersSent && result instanceof Promise) {
+				result.catch(next);
+			}
+		};
+	} else if (fn.length < 4) {
+		return (req, res, next) => {
+			const result = fn(req, res, next);
+			if (!result.headersSent && result instanceof Promise) {
```
i-like-robots: Perhaps this should explicitly throw a type error if the handler does not return either the response object or a promise?

maxcbc: Maybe. 1 idea I have had is to limit this to only support `async` functions and reject in `fn` isnt an `AsyncFunction`.

This would mean it wouldn't work for functions returning promises in general though.

src/lib/async-handlers.js
```diff
@@ -0,0 +1,25 @@
+exports.asyncHandler = fn => {
+	if (fn.length === 4) {
+		return (err, req, res, next) => {
+			const result = fn(err, req, res, next);
+			if (!result.headersSent && result instanceof Promise) {
+				result.catch(next);
+			}
+		};
```
i-like-robots: Could both conditional blocks be merged into one like this?:

```js
	return (...args) => {
		const result = fn.apply(null, args);

		if (!result.headersSent && result instanceof Promise) {
			result.catch(next);
		}
	};
```

PR #521, opened 2018-08-17
This bumps next metrics to 2.1.0. 

maxcbc:
This bumps next metrics to 2.1.0.  This version of next metrics includes a breaking change so we should probably release this as a major release.

 🐿 v2.10.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "ip": "^1.1.5",
     "isomorphic-fetch": "^2.2.1",
     "n-health": "^2.3.2",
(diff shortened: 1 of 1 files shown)


simonplend:
I don't think upgrading `next-metrics` to `v2.x.x` should be a major release as it's not a change in `n-express` itself.

I'm upgrading to `v2.2.x` and adding a health check for metrics configuration here: https://github.com/Financial-Times/n-express/pull/523

Are you happy to close this PR?

PR #541, opened 2018-11-14
Revert back to stable `next-metrics` v3.x releases

simonplend:
Related to #540.


simonplend:
Have realised there's no point reverting to the current stable release of `next-metrics` as it would mean adding the `app` option back in when we call `Metrics#init` from `n-express`. Instead I'll do a `next-metrics` stable minor release and roll the version forwards here.

PR #555, opened 2019-04-23
Updated automated n-express error rate healthchecks to use app.system…

richardLuko:
Currently n-express allows us to enter app.name and app.systemCode in its settings. We, in next-subscribe use both properties, and they are different. 
n-express also have few automated [health checks](https://github.com/Financial-Times/n-express/blob/e9b2962dc571a62341ee41092ec88faceff1639f/src/lib/error-rate-check.js#L17) which use app.name to generate Grafana path. 
This PR replaces usage of app.name with app.systemCode in error rate healthcheck

 🐿 v2.12.3

### DIFF
--- a/src/lib/guess-app-details.js
+++ b/src/lib/guess-app-details.js
@@ -19,6 +19,7 @@ module.exports = options => {
 	if (!name) throw new Error('Please specify an application name');
 
 	name = name && normalizeName(name);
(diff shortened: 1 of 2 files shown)


liamkeaton:
@richardLuko the tests were failing on this?

Issue #563, opened 2019-07-16
The error rate health check doesn't work


sjparkinson:
https://github.com/Financial-Times/n-express/blob/ef9a0dd4813c4898be33001df0b894298024aca9/src/lib/error-rate-check.js#L14-L24

From Bren during the recent stream page incident:

> looks like the `n-express` error response threshold healthcheck isn’t working
> because stream page is currently >60% `500` responses

We have seen success in more simple uses in next-health, https://github.com/Financial-Times/next-health/blob/master/server/config/health-checks/service-unavailability.js.

quarterto:
looks like, for `stream-page` at least, the graphite query uses the app name `next-stream-page`, but it's `stream-page` in the data.

this may be a problem with [`guessAppDetails`](https://github.com/Financial-Times/n-express/blob/master/src/lib/guess-app-details.js) (for `stream-page` or generally), but calling that function directly returns the right name.

this relates to Financial-Times/next#385

quarterto:
ah, it's using the `systemCode` not the name (as provided by the app itself). changing the `systemCode` to `stream-page` would fix this check for that app, but since the actual Biz Ops system code is `next-stream-page` i don't think this is the correct fix. i'm looking into how `next-metrics` puts things into Graphite and why it's a different name

quarterto:
jotting down ideas for future-Bren's benefit:

- we could do a combination `ebi` search/graphite call to check that there are no apps with a `systemCode` that doesn't exist in graphite
- it could become an `n-express` healthcheck! which checks graphite for an app with that system code.have to be careful because if we add it carelessly we'll suddenly get dozens of failing checks
- what else in `n-express` uses `systemCode`?

quarterto:
this is the complete list of services with a broken `errorThreshold` healthcheck

```
ads-api
ammit-api
ammit-docs
api
article
b2b-prospect
barrier-guru
consent-proxy
control-centre
corp-signup
epaper
es-interface
eventpromo-api
feedback-api
flag-toggler
flags-api
fow-api
front-page
graphics-page
health
houston
kat-api
list-doctor
live-event-api
lure-api
media-api
messaging-guru
myft-api
myft-page
myft-proxy
myft-slow-jobs
navigation-api
next-magnet-api
next-subscribe
personalised-feed-api
popular-api
preflight
preview
product
profile
retention
router
search-api
search-page
session
signup
signup-api
signup-proxy
spoor-consumer
spoor-device-atlas-api
[...]

PR #564, opened 2019-07-26
add support for separate graphiteName field

quarterto:
in case it's different to the systemCode (which it usually is). see #563. doesn't fix that issue on its own, but will once all 62 apps have a correct `graphiteCode`


sjparkinson:
Could you update `README.md` with details of the new option?

PR #585, opened 2020-05-27
Bump the next-metrics version

GlynnPhillips:


GlynnPhillips:
Yeah seemed to be timing/cache and its now resolved.

PR #588, opened 2020-07-27
Add withAb option to n-express so that res.locals.flags can optionally be set to include Ammit flag values

emmalewis:
n-flags-client has an option `withAb` which allows the values of flags to be set to their Ammit variant value (if the user is in a variant). This PR makes use of that and adds an optional `withAb` value to n-express so that apps can access these values if required. As things stand, in an app that's using n-express, an example flag `testAmmitFlag` with a specified `ammitSetup` and a `state` of `false` (flags with `ammitSetup` always have a `state` of `false` - see "Limitations and notes" here: https://github.com/Financial-Times/next-ammit-api#setup-details) we would see that `res.locals.flags.testAmmitFlag` would always equal `false`, because the state is false. If we were to release this PR and initialise n-express with `withAb` set to `true`, we would see that `res.locals.flags.testAmmitFlag` would switch between different possible variant values and false. This behaviour is desired in next-subscribe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Passed in to `nExpress`, these (Booleans defaulting to false unless otherwise st
 
 - `withBackendAuthentication` - Boolean, defaults to `true` - if there is a `FT_NEXT_BACKEND_KEY[_OLD]` env variable, the app will expect requests to have an equivalent `FT-Next-Backend-Key[-Old]` header; this turns off that functionality
 - `withFlags` - decorates each request with [flags](https://github.com/Financial-Times/n-flags-client) as `res.locals.flags`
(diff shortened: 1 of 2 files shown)


andygout:
Just had a quick chat with @nickramsbottom offline and he has clarified some points:

There is a distinction between flags (on/off) and Ammit flags (control/variant(s)). While flags get attached to the header, the Ammit flags do not and need to be retrieved by the app (which happens in the likes of `next-subscribe` and [`next-barrier-guru`](https://github.com/Financial-Times/next-barrier-guru/blob/master/server/lib/ammit.js)).

This change will provide the option for apps to acquire Ammit flags far more easily and enable code re-use (and resultantly cleanup of current code that repeats the same task).

emmalewis:
> Would it be possible to add this option to the README as well?

@nickramsbottom Thank you - of course. That's done now.

nickramsbottom:
Great stuff! :shipit:

PR #595, opened 2020-09-25
change CMDB reference to Biz Ops

apaleslimghost:


nayanashettyft (approved):

nayanashettyft (commented):
sorry missed checking the test run status before approving the PR. Looks like https://github.com/Financial-Times/n-express/blob/master/test/app/app.test.js#L169 would need updating as well

apaleslimghost:
ah, good spot, updated it

PR #611, opened 2020-12-14
Rename whitelist -> allowlist

andygout:
Renames 'whitelist' -> 'allowlist' in line with [TGG: Address offensive and objectionable technology terms](https://docs.google.com/document/d/1v6z7_NkLFeYAFotTYB8C1OguO5iMHHLpAOv_yG5EKHI/edit#).

From what I can see, the only change we should expect outside of this package is that logged event values will change from (e.g.) `IP_WHITELIST_UPDATE` to `IP_ALLOWLIST_UPDATE`, so we should be aware of that mixture during the migration process.

I think a patch release is probably appropriate - would others agree?


apaleslimghost (approved):
👍 for the change, but i can't work out where, if anywhere, this is used? it doesn't seem to be included anywhere else in `n-express`, or by any apps? if it were being used the name change should be a major change, but can we just delet it instead?

andygout:
Agree - I can't see that a module is exported under the name `ipWhitelist`: https://github.com/search?q=user%3AFinancial-Times+ipWhitelist&type=Code.

I will open a separate PR to delete the dead code.

andygout:
Superseded by https://github.com/Financial-Times/n-express/pull/612.

PR #612, opened 2020-12-14
Remove unused IP whitelist code

andygout:
The original intention was just to rename 'whitelist' -> 'allowlist' in line with [TGG: Address offensive and objectionable technology terms](https://docs.google.com/document/d/1v6z7_NkLFeYAFotTYB8C1OguO5iMHHLpAOv_yG5EKHI/edit#).

However, as per https://github.com/Financial-Times/n-express/pull/611, it transpires that the `ip-whitelist` module is not being employed whatsoever (https://github.com/search?q=user%3AFinancial-Times+ipWhitelist&type=Code) meaning we are in a position to remove the dead code, which is done in this PR.


andygout:
Would others agree with a patch release?

apaleslimghost (approved):
did a bit more digging, looks like this used to be used as a fallback for backend keys, but it was removed three days after it was added back in 2017 and the code was never cleaned up https://github.com/Financial-Times/n-express/commit/453d0a11eba2f66e7bd60410b6595aa2b3d97558

andygout:
Nice detective work.

Three and a half years without using it seems like sufficient assurance.

apaleslimghost:
oh, just realised you should remove the `ip` package from `package.json` as well

apaleslimghost:
nooooo one second too late

PR #628, opened 2021-04-23
light refactor of a couple of files

apaleslimghost:
split out from my CPP-431 branch

### DIFF
--- a/src/lib/instrument-listen.js
+++ b/src/lib/instrument-listen.js
@@ -49,7 +49,7 @@ module.exports = (app, meta, initPromises) => {
 			nLogger.warn({ event: 'EXPRESS_START', app: meta.name, port: port, nodeVersion: process.version });
 
 			if(callback) {
-				return cb.apply(this, arguments);
+				return callback.apply(this, arguments);
 			}
 		};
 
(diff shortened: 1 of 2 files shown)


leannecornish-ft (approved):

src/lib/instrument-listen.js
```diff
@@ -5,57 +5,66 @@ const http = require('http');
...
+		} catch(err) {
+			// crash app if initPromises fail by throwing an error asynchronously outside of the promise
+			// TODO: better error handling
+			setTimeout(() => {
 				throw err;
-			}));
+			});
+		}
+
+		metrics.count('express.start');
```
kavanagh, line 66: Would it be better if this line lived in `wrappedCallback` along side the similar log message? As that's more accurately when express has started.

I'm not sure what the benefit of having both is TBH.

apaleslimghost: `express.start` isn't a log, it's for Graphite, our Grafana dashboards mark it on the time series. i deliberately haven't changed where this is happening in this PR because it's just a refactor. this metric has been counted in the exact same place [literally since 2015](https://github.com/Financial-Times/n-express/commit/a7612cc00d6bf9a17a898d166f66b5b6f4ec2b29), if we want to change it i think it should be more considered.

kavanagh: Of course I know the difference 😄 . I get it's just a refactor. It's just that putting this on line 50 makes the two things that make the same event happen in the same place.

apaleslimghost: if you think that's better, go ahead and raise a PR. it's been fine where it is for six years is all

PR #629, opened 2021-04-23
no hardcoded certificates

apaleslimghost:
- load certicates from the app's directory
- add a script to generate certificates 4 u

this is a breaking change


src/lib/instrument-listen.js
```diff
@@ -11,24 +11,20 @@ const readFile = denodeify(fs.readFile);
...
+				throw Error('n-express was started with --https, but there\'s no self-signed certificate or key in your app directory. run `npx n-express-generate-certificate` to create one')
+			});
 
 			return https.createServer({ key, cert }, app);
 		} else {
 			return http.createServer(app);
 		}
 	}
 
-	async function actualAppListen(app, ...args) {
```
emortong, line 26: trying to understand why this `actualAppListen` function was being used in the first place, it looks that what you changed it for is doing the same thing, but I am wondering if I am missing something.

emortong: ahh I had not seen that the function was being assigned to `app.listen` on line 32, thanks!

emortong (approved):

PR #640, opened 2021-07-14
Roll back v22 breaking changes

leannecornish-ft:
This is to roll back the v22 changes that were added to v21 accidentally.


leannecornish-ft:
I think a rebase and merge would be best for this - please confirm.

ivomurrell:
> I think a rebase and merge would be best for this - please confirm.

Yes, we probably want to rebase your changes onto the `v21` branch rather than having them based off of `main`?

apaleslimghost (approved):
just checking i've got the right end of the stick, this should be v21, with the recent changes, minus the v22 breaking changes? if that's the case, looks good

leannecornish-ft:
> just checking i've got the right end of the stick, this should be v21, with the recent changes, minus the v22 breaking changes? if that's the case, looks good

Yes, exactly right. Thanks!

PR #642, opened 2021-07-16
Add type declarations

### LINKED ISSUE #637: Export express types
Hello Platforms team!

We are migrating `next-discovery-pages` to Typescript and we found a small issue when trying to type the Request/Response objects. Basically we are required to install / import them from `@types/express` and do some `@ts-ignore` to fix the fact that, for example the response object have extra properties:

https://github.com/Financial-Times/next-discovery-pages/pull/455/files

```
	const headers = {
		// @ts-ignore FT_NO_CACHE not available in res
		'Cache-Control': page.cacheControl || res.FT_NO_CACHE,
		// @ts-ignore FT_DAY_CACHE not available in res
		'Surrogate-Control': page.surrogateControl || res.FT_DAY_CACHE,
		'Surrogate-Key': surrogateKeys.join(' '),
	}
```

We were wondering if it would be possible that `n-express` exports the extended express types so the migration is more straightforward.

ivomurrell:
This PR adds type declarations to `n-express`, both for packages using TypeScript to consume, as well as to enforce typing within the project via JSDoc comments. It might be worth considering moving the declaration files onto `DefinitelyTyped` in the future, but as a preliminary phase it's fine to bundle them with the package.

Tested the types with the `next-discovery-pages` repository, though it might be worth adding unit tests in the future.

Thanks to @oliverturner for the original spike!

Closes #637


main.js
```diff
@@ -23,30 +28,44 @@ const healthChecks = require('./src/lib/health-checks');
...
-		healthChecks: []
-	}, options || {});
+const teapot = fs.readFileSync(
+	path.join(__dirname, 'src/assets/teapot.ascii'),
+	'utf8'
+);
+
+/**
+ * @param {AppOptions} options
+ * @returns {import("./typings/n-express").AppContainer}
```
apaleslimghost, line 44: **nitpick**: for consistency, and maybe it'll read cleaner, it would be good to have these imports at the top of the file with the others, not inline

typings/n-express.d.ts
```diff
@@ -0,0 +1,80 @@
...
+	withBackendAuthentication: boolean;
+	withFlags: boolean;
+	withServiceMetrics: boolean;
+}
+
+export interface AppError {
+	severity?: number;
+	threshold?: number;
+	samplePeriod?: string;
+}
```
apaleslimghost, line 53: **suggestion**: the name `AppError` for this is potentially misleading, might imply it's an actual Error object. `ErrorRateHealthcheckOptions` would be clearer

src/middleware/cache.js
```diff
@@ -1,10 +1,29 @@
...
+
+const cacheHeaders = {
+	FT_NO_CACHE: 'max-age=0, no-cache, must-revalidate',
+	FT_NO_CACHE_PRIVATE: 'max-age=0, no-cache, no-store, must-revalidate, private',
+	FT_SHORT_CACHE: 'max-age=600, stale-while-revalidate=60, stale-if-error=86400',
+	FT_HOUR_CACHE: 'max-age=3600, stale-while-revalidate=60, stale-if-error=86400',
+	FT_DAY_CACHE: 'max-age=86400, stale-while-revalidate=60, stale-if-error=86400',
+	FT_WEEK_CACHE: 'max-age=604800, stale-while-revalidate=60, stale-if-error=259200',
+	FT_LONG_CACHE: 'max-age=86400, stale-while-revalidate=60, stale-if-error=259200'
+};
```
apaleslimghost, line 13: **thought**: i hate that these are just attached to `res` for no reason other than it's there. i'd love to eventually see these as exports instead, and replace these with getters that output a deprecation warning when used (probably using `util.deprecate`)

typings/n-express.d.ts
```diff
@@ -0,0 +1,80 @@
...
+	unVaryAll: unvaryAll;
+}
+
+export interface NextFunction extends Express.NextFunction {}
+
+export type Callback = (
+	req: Request,
+	res: Response,
+	next: NextFunction
+) => void;
```
apaleslimghost, line 27: **thought**: it doesn't look like this type is used for express's `app.get` etc, so apps will only see the base express `Request` etc there, do you plan to address that?

apaleslimghost: oh i guess that's what the TODO comment below is talking about

src/middleware/backend-authentication.js
```diff
@@ -14,43 +24,61 @@ module.exports = (app, appName) => {
...
+			req,
+			res,
+			next
+		) => {
+			// TODO - change how all this works in order to use __assets/app/{appname}
+			// allow static assets, healthchecks, etc., through
+			if (
+				req.path.indexOf('/' + appName) === 0 ||
+				req.path.indexOf('/__') === 0
+			) {
```
apaleslimghost, line 52: **praise**: these formatting changes make it _much_ easier to read

src/middleware/backend-authentication.js
```diff
@@ -14,43 +24,61 @@ module.exports = (app, appName) => {
 	if (!backendKeys.length) {
 		const missingEnvVarsMessage = [
 			'Backend authentication is disabled due to missing environment variables -',
-			'If backend authentication isn\'t required then disable this middleware by setting the n-express \'withBackendAuthentication\' option to false.',
+			"If backend authentication isn't required then disable this middleware by setting the n-express 'withBackendAuthentication' option to false.",
 			'To enable backend authentication add the missing environment variables.',
 			'For more info on backend authentication see https://github.com/Financial-Times/n-express#optional',
 		];
```
apaleslimghost, line 30: **suggestion**: why not go further and make the whole thing a template string

src/lib/instrument-listen.js
```diff
@@ -33,14 +35,14 @@ module.exports = (app, meta, initPromises) => {
 		app.use(raven.errorHandler());
 
 		// Optional fallthrough error handler
-		app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
+		app.use(/** @type {} */(err, req, res, next) => { // eslint-disable-line no-unused-vars
```
apaleslimghost, line 38: **question**: i wouldn't expect this callback to be cast to an empty object here, is there a reason for it?

apaleslimghost (approved):
looks great! a few non-blocking comments

main.js
ivomurrell: My logic was to put the imports inline if the type was only used once, and at the top if it was reused, but consistency probably is more readable!

apaleslimghost: ah! that probably does make sense tbh. up to you!

typings/n-express.d.ts
ivomurrell: > oh i guess that's what the TODO comment below is talking about

Yeah exactly, I was hoping there would be some way to map the function signatures to use our `NExpress` type instead but no dice as far as I can tell. Will probably have to copy and paste the `express` definition instead but that's a bit messy and fragile so I think we should put it off until we see a case where it's actually needed first.

ivomurrell:
Seeing as this PR is based on an experimental spike there are a few changes (quite useful ones imo) included that don't really relate to the final PR's topic, e.g., prettier configuration + Volta node version pinning. Would it make sense to split those off into their own separate PRs instead?

src/middleware/backend-authentication.js
ivomurrell: This was a Prettier change tbf, not really related to this PR's topic. See the comment I made below

apaleslimghost:
> Would it make sense to split those off into their own separate PRs instead?

good idea, but i'm happy either way

PR #646, opened 2021-07-29
Remove withAb option

jamesnicholls:
As of v8.7.1, n-flags-client no longer takes a `withAb` option. See also https://github.com/Financial-Times/n-flags-client/pull/157.


package-lock.json
```diff
@@ -1,8 +1,14126 @@
 {
   "name": "@financial-times/n-express",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
apaleslimghost, line 4: **issue**: this lockfile was generated with npm 7, which we don't fully support yet. could you revert this change please?

main.js
```diff
@@ -51,7 +51,6 @@ const getAppContainer = (options) => {
 		{
 			withBackendAuthentication: true,
 			withFlags: false,
-			withAb: false,
```
apaleslimghost, line 54: **suggestion (non-blocking)**: could we add a deprecation notice so that apps still passing in this option get prompted to remove it?

apaleslimghost (changes requested):
thanks @jamesnicholls! one issue to address, and one non-blocking suggestion

main.js
jamesnicholls: Yep, good idea, will add.

package-lock.json
jamesnicholls: Whoops, will remove.

apaleslimghost (approved):
fantastic, thank you!

PR #652, opened 2021-08-26
CPP-555 pass system code to n-logger

apaleslimghost:
depends on a version of n-logger that hasn't been released yet, so CI will fail until it is


apaleslimghost:
not going ahead with this.

PR #676, opened 2022-01-11
FTDCS-37 Fix the restart check tests

rowanmanning:
This addresses the test failures introduced in #675, which were caused
because we were expecting only a single health check to be exported by
the `health-checks` module.

After a discussion with @serena97 and @ivomurrell, we decided to
refactor the `health-checks` module to export and array of checks,
and update `InstrumentListen#addMetrics` to accept either an array
or a single item.

Let me know if anything isn't quite right, happy to make any code
style changes too (it's been a while!)


ivomurrell (approved):
Looks great! Thanks for taking the time to write some docs and improve the tests too)

PR #677, opened 2022-01-26
[FTDCS-37] Update app restart check

alexmuller:
Commits:

- Change the metric that we're testing so that it's based on region
- Use a graphite threshold instead of spike - for consistency/reliability

The way the `graphiteThreshold` check works is a little confusing at first but very good for us:

- The wildcard matcher (for example `*_EU`) will match all dynos for a given app
- Graphite will return multiple series for this query (`dyno_1_EU`, `dyno_2_EU`, etc)
- The graphiteThreshold check will iterate over all these series and check if any of them are above a given value
- I have used the `summarize()` function to group these series into 1 hour buckets, and `transformNull()` to set null data (no restarts) to 0
- The samplePeriod means that we go back 6 hours (6 metrics per dyno)

I reckon if a dyno restarted more than 2 times in a one hour period that would be a good indication of a problem.

### DIFF
--- a/src/lib/app-restart-check.js
+++ b/src/lib/app-restart-check.js
@@ -16,22 +16,22 @@ module.exports = (appName, opts) => {
 	opts = opts || {};
 	const severity = opts.severity || DEFAULT_SEVERITY;
 
-	let region_name, dyno_matcher;
+	let regionName, dynoMatcher;
 
 	if (process.env.REGION) {
-		region_name = process.env.REGION.toUpperCase();
-		dyno_matcher = `*_${region_name}`;
+		regionName = process.env.REGION.toUpperCase();
+		dynoMatcher = `*_${regionName}`;
 	} else {
-		region_name = 'unknown region';
-		dyno_matcher = '*';
+		regionName = 'unknown region';
+		dynoMatcher = '*';
 	}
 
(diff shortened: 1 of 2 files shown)


src/lib/app-restart-check.js
```diff
@@ -16,20 +15,30 @@ const DEFAULT_THRESHOLD = 20;
...
-			type: 'graphiteSpike',
-			threshold: threshold,
-			baselinePeriod: '14d',
-			samplePeriod: '1hour',
-			numerator: `next.heroku.${appName}.*.express.start`,
+			name: `${appName} restart rate is normal (${region_name})`,
+			type: 'graphiteThreshold',
+			threshold: 2,
+			samplePeriod: '6hours',
+			metric: `transformNull(summarize(${applicationStartMetric}, '1h'), 0)`,
```
apaleslimghost, line 38: TIL `transformNull`! that's very useful

apaleslimghost (approved):
this looks great! will be fantastic to have this check be more reliable

rowanmanning (approved):
This looks great 👍  very useful.

I went looking for tests to double-check you haven't missed any required changes, but they don't exist. I don't really want to block this work but some tests for the region name and dyno matching logic might be good at some point

rowanmanning (approved):
Tests look good to me 🙂 👍

PR #699, opened 2022-05-17
Add Possibility to use node-fetch instead isomorphic fetch

juanSanchezAlcala:
Node-fetch is been used in some projects due it has more capabilities than isomorphic fetch. This causes that the additonal modifications that we do the fetch library like instrumenting with next-metrics is not been done, so metrics are not send .

Due to Isomorphic fetch modifies global.fetch we are setting in this variable the node-fetch if useNodeFetch is called in order to not affect other apps that use n-express

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,7 +87,7 @@ jobs:
             .circleci/shared-helpers
       - *restore_npm_cache
       - node/install-npm:
-          version: "7"
+          version: "7.20.2"
       - run:
           name: Install project dependencies
           command: make install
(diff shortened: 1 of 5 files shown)


apaleslimghost (commented):
hi @juanSanchezAlcala, thanks for this PR. if an app wants to use `node-fetch`, it can always `require` it directly and ignore the global `fetch` from `isomorphic-fetch`.

i would rather not be setting `global.fetch` in a library (the decision to use `isomorphic-fetch` was made long before my team owned this library, otherwise we wouldn't have made that choice), so we won't be merging this.

since node 18 now has native `fetch`, we'll probably actually remove `isomorphic-fetch` in a PR instead.

PR #701, opened 2022-05-24
FTDCS-188 - Add system code the debug headers

JSRedondo:
We need to associate Fastly logs with systems. 
This currently isn’t possible - the worst example is next-stream-page (in Fastly logs, there’s no way to figure out that /world is served by next-stream-page.
We should surface this data in Fastly so that we can group errors by system.


apaleslimghost (approved):

alexmuller (approved):
This looks good to me, thanks!

PR #702, opened 2022-05-30
Add express text function to module.exports

JSRedondo:
Add [express text function](https://expressjs.com/en/api.html#express.text) to module.exports.

We can eliminate body-parser dependency in all the repositories using json and/or text functions.


ivomurrell (approved):
🙌🏻

PR #711, opened 2022-07-14
[FTDCS-214] Remove restart rate default health check

alexmuller:
This check is no longer useful - it was to try and tell us which apps were crashing frequently but we had too many false positives (for example if you deploy your app many times in a day it would go off).

We have better monitoring in next-health now - specifically for app crashes - so this check can go.

Also I had to update some strange tests because this is the only default severity 2 alert.


rowanmanning (approved):
Nice 👍 I remember looking through these tests with Ania and finding it confusing

PR #712, opened 2022-07-15
Upgrade next-metrics from v7.0.0 to v7.1.0

alexmuller:
This includes the Heroku Metrics API in services.js which we need for next-health.


rowanmanning (approved):
Looks good to me 👍

PR #714, opened 2022-07-21
Add a default Heroku log drain health check

rowanmanning:
This uses the new Heroku log drain check which allows us to see alerts
for an app if it's expecting to use log drains but they're misconfigured
in some way. This default check is only enabled if an app has the
`MIGRATE_TO_HEROKU_LOG_DRAINS` environment variable set so it won't
start failing for apps which haven't conciously opted into log drains
(See https://github.com/Financial-Times/n-logger/pull/152 for more info).

There's a note in the source explaining that this condition will be
removed in future. I think this is a minor change as the health check
will only start failing when an app opts in.

Relates-To: FTDCS-218

### DIFF
--- a/src/lib/heroku-log-drain-check.js
+++ b/src/lib/heroku-log-drain-check.js
@@ -16,7 +16,7 @@ module.exports = (opts) => {
 
 	return nHealth.runCheck({
 		id: `heroku-log-drain-${region}`,
-		name: 'Heroku log drain configured',
+		name: `Heroku log drain configured (${region.toUpperCase()})`,
 		type: 'herokuLogDrain',
 		severity: opts.severity
 	});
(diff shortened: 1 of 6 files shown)


src/lib/heroku-log-drain-check.js
```diff
@@ -0,0 +1,23 @@
...
+ * @returns {Healthcheck & TickingMetric}
+ */
+module.exports = (opts) => {
+	opts = opts || {};
+
+	const region = process.env.REGION === 'US' ? 'us' : 'eu';
+
+	return nHealth.runCheck({
+		id: `heroku-log-drain-${region}`,
+		name: 'Heroku log drain configured',
```
alexmuller, line 19: __nitpick:__ Would you be happy to have something in the name which differentiates the checks per region? The app name would be cool but I don't think we have that here. So maybe just the region.

Otherwise we end up with multiple checks in Heimdall with the same name which are hard to distinguish.

alexmuller (approved):
🌟

PR #718, opened 2022-08-02
Add supported Node.js version healthcheck

andygout:
FTDCS-255

Jira card: https://financialtimes.atlassian.net/jira/software/c/projects/FTDCS/boards/1426?modal=detail&selectedIssue=FTDCS-255

> We have apps running unsupported versions of Node.js and we need to make sure that they don’t get forgotten about when we’re upgrading. For example next-barrier-guru is running Node 12. We also need to make sure that all our Node 14 apps are updated before next April.

> #### Definition of done

> - We’ve added a new healthcheck type to n-health which checks the Node version and applies some logic to fail the check:
>    - We should only run long term support (LTS) versions of Node
>    - 12 or older should fail because they’re unsupported
>    - 14/16/18 have future dates at which they should start to fail:  

> - Add this new check type to n-express as a default check so that it’s automatically rolled out to every app

The added healthcheck looks like this:
Screenshot 2022-08-02 at 10 27 01

#### Code decision
I have chosen not to re-perform the check upon each call to `getStatus()`; the Node.js version can only be changed via the app re-starting so it makes sense to only run the check once when the app starts and then return the same value until the dynos next restart. Consequently, the `lastUpdated` value will reflect the time of the app starting (and the healthcheck being run this one time) and not update thereafter.

#### Testing decision
Ideally the unit tests would be able to contrive a context whereby the `ok` and `checkOutput` for various Node.js versions could be tested, but overriding the `process.versions.node` value (whether even possible) would result in the updated Node.js version being the version on which the tests are actually running, resulting in an inconsistent test environment (both locally and in CI where Node.js version images are specified) that could be the cause of low-level issues that are not worth the effort that the unit tests provide. So I have gone with some more basic tests that ensure a healthcheck response is received with the region assigned as expected.

#### Release version
I plan to issue this change as a minor release as it adds functionality in a backwards compatible manner.

### DIFF
--- a/src/lib/health-checks.js
+++ b/src/lib/health-checks.js
@@ -5,11 +5,14 @@
  * @typedef {import("../../typings/n-express").AppOptions} AppOptions
  */
 
-const errorRateCheck = require('./error-rate-check');
-const unRegisteredServicesHealthCheck = require('./unregistered-services-healthCheck');
-const metricsHealthCheck = require('./metrics-healthcheck');
-const herokuLogDrainCheck = require('./heroku-log-drain-check');
 const nLogger = require('@financial-times/n-logger').default;
+
+const errorRateCheck = require('./error-rate-check');
+const herokuLogDrainCheck = require('./heroku-log-drain-check');
+const metricsHealthCheck = require('./metrics-healthcheck');
+const supportedNodeJsVersionCheck = require('./supported-node-js-version-check');
+const unRegisteredServicesHealthCheck = require('./unregistered-services-healthCheck');
+
 /**
  * @param {ExpressApp} app
  * @param {AppOptions} options
@@ -39,7 +42,8 @@ module.exports = (app, options, meta) => {
 	const defaultChecks = [
 		...tickingMetricChecks,
 		unRegisteredServicesHealthCheck.setAppName(meta.name),
-		metricsHealthCheck(meta.name)
+		metricsHealthCheck(meta.name),
+		supportedNodeJsVersionCheck(meta.name)
 	];
 
 	// See https://github.com/Financial-Times/n-logger#loggers for information on
--- a/src/lib/supported-node-js-version-check.js
+++ b/src/lib/supported-node-js-version-check.js
@@ -1,0 +1,46 @@
+const semver = require('semver');
+
+// Ref. https://nodejs.org/en/about/releases
+const nodeVersionToEndOfLifeDateMap = {
+	'12': new Date('2022-04-30'),
+	'14': new Date('2023-04-30'),
+	'16': new Date('2023-09-11'),
+	'18': new Date('2025-04-30'),
+	'20': new Date('2026-04-30')
+};
+
+/**
+ * @param {string} appName
+ * @returns {import("../../typings/metrics").Healthcheck}
+ */
+module.exports = (appName) => {
+	const lastUpdated = new Date();
+
+	const currentDate = new Date();
+
+	const region = process.env.REGION === 'US' ? 'us' : 'eu';
+
+	const nodeVersion = process.versions.node;
+
+	const nodeMajorVersion = semver.major(nodeVersion);
+
+	const hasSupportedVersion = nodeVersionToEndOfLifeDateMap[nodeMajorVersion] > currentDate;
+
+	return {
+		getStatus: () => {
+			return {
+				id: `supported-node-js-version-${region}`,
+				name: `Long-term supported (LTS) Node.js version used for ${appName} (${region.toUpperCase()})`,
+				ok: hasSupportedVersion,
+				checkOutput: hasSupportedVersion
+					? `${appName} uses v${nodeVersion} of Node.js, which is a long-term supported (LTS) version`
+					: `${appName} uses v${nodeVersion} of Node.js, which is not a long-term supported (LTS) version`,
+				lastUpdated,
+				severity: 3,
+				businessImpact: 'Systems may be subject to security and/or functional bugs that will not be fixed and which could manifest in various ways.',
+				technicalSummary: 'Inspects the version of Node.js running the application and verifies that the end-of-life date for it has not passed.',
+				panicGuide: 'Upgrade the app to a long-term supported version of Node.js at your earliest convenience.'
+			};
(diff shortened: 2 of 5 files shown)


src/lib/health-checks.js
```diff
@@ -5,11 +5,14 @@
...
+const nLogger = require('@financial-times/n-logger').default;
+
 const errorRateCheck = require('./error-rate-check');
-const unRegisteredServicesHealthCheck = require('./unregistered-services-healthCheck');
-const metricsHealthCheck = require('./metrics-healthcheck');
 const herokuLogDrainCheck = require('./heroku-log-drain-check');
-const nLogger = require('@financial-times/n-logger').default;
+const metricsHealthCheck = require('./metrics-healthcheck');
+const supportedNodeJsVersionCheck = require('./supported-node-js-version-check');
+const unRegisteredServicesHealthCheck = require('./unregistered-services-healthCheck');
```
andygout, line 14: The net change here is the addition of `const supportedNodeJsVersionCheck = require('./supported-node-js-version-check');`; otherwise it is separating external dependencies from native ones, and alphabetising the native dependencies.

src/lib/supported-node-js-version-check.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+/**
+ * @param {string} appName
+ * @returns {Healthcheck & TickingMetric}
+ */
+module.exports = (appName) => {
+	const lastUpdated = new Date();
+
+	const currentDate = new Date();
```
andygout, line 17: I tried to use `const lastUpdated = currentDate = new Date();` but it failed the linting rules (`no-undef`).

src/lib/health-checks.js
andygout, line 14: The net change here is simply the addition of `const supportedNodeJsVersionCheck = require('./supported-node-js-version-check');`; otherwise it is separating external dependencies from native ones, and alphabetising the native dependencies.

src/lib/supported-node-js-version-check.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+/**
+ * @param {string} appName
+ * @returns {import("../../typings/metrics").Healthcheck}
+ */
+module.exports = (appName) => {
+	const lastUpdated = new Date();
+
+	const currentDate = new Date();
```
andygout, line 17: I tried to use `const lastUpdated = currentDate = new Date();` here but it failed the linting rules (`no-undef`).

src/lib/supported-node-js-version-check.js
```diff
@@ -0,0 +1,46 @@
...
+
+	const region = process.env.REGION === 'US' ? 'us' : 'eu';
+
+	const nodeVersion = process.versions.node;
+
+	const nodeMajorVersionRegex = new RegExp(/^([0-9]*)\./);
+
+	const nodeMajorVersion = nodeVersion.match(nodeMajorVersionRegex)[1];
```
rowanmanning, line 25: **suggestion:** I'd be tempted to use the [semver](https://www.npmjs.com/package/semver) package here rather than a regular expression. Not essential and non-blocking, but it'd be easier to read and less error-prone.

This module is already included a bunch of times in `package-lock.json` so it wouldn't increase the bundle size or anything if we installed it as a regular dependency.

```suggestion
	const nodeVersion = process.versions.node;
	const nodeMajorVersion = semver.major(nodeVersion);
```

src/lib/supported-node-js-version-check.js
```diff
@@ -0,0 +1,46 @@
...
+				id: `supported-node-js-version-${region}`,
+				name: `Long-term supported (LTS) Node.js version used for ${appName} (${region.toUpperCase()})`,
+				ok: hasSupportedVersion,
+				checkOutput: hasSupportedVersion
+					? `${appName} uses v${nodeVersion} of Node.js, which is a long-term supported (LTS) version`
+					: `${appName} uses v${nodeVersion} of Node.js, which is not a long-term supported (LTS) version`,
+				lastUpdated,
+				severity: 3,
+				businessImpact: 'Systems may be subject to security and/or functional bugs that will not be fixed and which could manifest in various ways.',
+				technicalSummary:`${appName} is not using a long-term supported version of Node.js and so will not receive critical bug fixes.`,
```
rowanmanning, line 41: **suggestion:** I think technical summary kind of duplicates the business impact here, [technical summary](https://docs.google.com/document/d/1ftlkDj1SUXvKvKJGvoMoF1GnSUInCNPnNGomqTpJaFk/edit#heading=h.ub010a1bgfy8) should explain what the check is testing I think, so maybe something like:

```suggestion
				technicalSummary: 'Inspects the version of Node.js running the application and verifies that the end-of-life date for it has not passed',
```

rowanmanning (changes requested):
Couple of little things from me. I think the technical summary should be changed but I'm easy if you don't want to switch to the semver package.

alexmuller (commented):
This is looking great - nice and self-contained. Works perfectly for me in the `test/fixtures/` app with different Node versions.

💯 for what you've written about how we don't need this check to tick. The app will have to restart to pick up a new Node version. It has made me realise that we have a few different ways to schedule check updates which is a little confusing. But not something to fix now.

If the Node version isn't in the object we'll compare `undefined > currentDate` which is false 👍 

Dates look right to me 👍

src/lib/supported-node-js-version-check.js
andygout: This is much cleaner - I have applied this change in commit [Add semver package to extract major version from Node.js version](https://github.com/Financial-Times/n-express/pull/718/commits/97f6b949446e28c402a4af27223c1a89b8a6ee63)

src/lib/supported-node-js-version-check.js
andygout: 👍 Updated in commit [Add supported Node.js version healthcheck](https://github.com/Financial-Times/n-express/pull/718/commits/47faf2ec52439dbe003d678639c59217249aa266).

rowanmanning (approved):
Looks great 👍 thanks for making those tweaks!

PR #719, opened 2022-08-09
Upgrade n-health ^7.0.0 -> ^7.0.1

andygout:
Consumes a minimum `n-health` version of [v7.0.1](https://github.com/Financial-Times/n-health/releases/tag/v7.0.1) so that we can state this as a minimum version requirement as part of the [Migrating an app to Heroku log drains](https://financialtimes.atlassian.net/wiki/spaces/DS/pages/7883555001/Migrating+an+app+to+Heroku+log+drains) guide.

### Release version
Like the corresponding `n-health` release version, this should also be a patch release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "denodeify": "^1.2.1",
         "express": "^4.17.3",
         "isomorphic-fetch": "^3.0.0",
-        "n-health": "^7.0.0",
+        "n-health": "^7.0.1",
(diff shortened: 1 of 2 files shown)


rowanmanning (changes requested):
This also needs a bump in `package-lock.json`, just running `npm i` should do it

andygout:
Of course! 🤦

rowanmanning (approved):
Great 🙂

PR #720, opened 2022-08-17
ACC-1874  bump up next-metrics

sakibshaik:
Description:
Updating next-metrics to 7.2.0

Ticket:
https://financialtimes.atlassian.net/browse/ACC-1874


joelcarr (changes requested):
Please can you reflect this change in the `package-lock.json` as well?

joelcarr (approved):

PR #727, opened 2022-09-14
Upgrade n-flags-client ^11.1.0 -> ^12.0.0; n-health ^7.0.2 -> ^8.0.0

andygout:
This PR is in relation to https://github.com/Financial-Times/next-preflight/pull/645, which resulted in `next-preflight` consuming a mixture of `isomorphic-fetch` v2 (which consumes `node-fetch` v1) and v3 (which consumes `node-fetch` v2).

The working assumption is that this mixture of `node-fetch` versions had been present before now but that the specific composition and ordering of dependencies meant that an instance of `node-fetch` v1 was always identified first and consequently the version that was employed by the code but that the changes in the above-mentioned PR meant that `node-fetch` v2 was the version employed by the code and that `next-preflight` had not yet been updated to accommodate any major breaking changes.

As a result, this caused [`next-preflight/server/tasks/next/consent.js` l.46](https://github.com/Financial-Times/next-preflight/blob/main/server/tasks/next/consent.js#L46) to throw:

`TypeError: Cannot convert undefined or null to object`

See Slack incident channel **#inc-1750-my-account-page-is-missing-tabs** for more details.

This error manifested itself in the FT.com UI by removing navigation in some pages (first observed on https://www.ft.com/myaccount/personal-details) owing to `next-navigation-api` returning 404 errors (a side effect we are still diagnosing).

This PR upgrades this package's:
- `n-flags-client` version to v12 (which in turn consumes `n-eager-fetch` v6, which in turn consumes `isomorphic-fetch` v3, which in turn consumes `node-fetch` v2)
- `n-health` version to v7 (which in turn consumes `node-fetch` v2)

These are steps to creating the availability of all packages with a `node-fetch` dependency or sub-dependency consumed by `next-preflight` to guarantee that `node-fetch` v2 will be the version employed by the code.

### Release
Given how the mixture of `node-fetch` versions can affect an app as demonstrated in the above incident, I propose issuing this change as a **major** release so that consumers are more conscious of the change they are making.

### References
- [GitHub: node-fetch / node-fetch - Upgrade to node-fetch v2.x](https://github.com/node-fetch/node-fetch/blob/main/docs/v2-UPGRADE-GUIDE.md)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,14 +9,14 @@
       "version": "0.0.0",
       "hasInstallScript": true,
       "dependencies": {
-        "@financial-times/n-flags-client": "^11.1.0",
+        "@financial-times/n-flags-client": "^12.0.0",
         "@financial-times/n-logger": "^10.2.0",
         "@financial-times/n-raven": "^6.3.0",
         "debounce": "^1.1.0",
         "denodeify": "^1.2.1",
         "express": "^4.17.3",
         "isomorphic-fetch": "^3.0.0",
-        "n-health": "^7.0.2",
+        "n-health": "^8.0.0",
         "next-metrics": "^7.3.0",
         "semver": "^7.3.7"
       },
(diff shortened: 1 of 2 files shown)


andygout:
I've eyeballed for uses of `fetch` here that would break in `node-fetch` v2 (based on the [`node-fetch` upgrade doc](https://github.com/node-fetch/node-fetch/blob/main/docs/v2-UPGRADE-GUIDE.md)) and cannot see anything problematic.

ivomurrell (approved):
yeah it looks like it just adds `fetch` to the global scope here, rather than using the API at all 👍🏻

PR #732, opened 2022-09-26
Remove the Heroku Log Drain health check

rowanmanning:
We decided that we're going to move this check to next-health. This is because we didn't consider the implications of accessing the Heroku API for apps which have many dynos or use Node.js clustering.

We're fine with removing this before we have next-health set up properly, because:

  1. The risk is low that an already-migrated app will have the log drain re-configured during this time

  2. We need this to be released in order to finish migrating next-preflight. Right now Preflight is in an intermediate state which is not ideal, and the risk of us having badly formatted logs outweighs the risk of log drains being re-configured across other applications

Despite the removal of a health check, I'd like to release this as a patch version. My thinking is that an app will not slip into a negative state because of this removal, whereas _adding_ a new health check has a chance of making that app alert unexpectedly.

It could also be considered a "fix" because we clearly didn't think this check through originally and it caused apps to break.


ivomurrell (approved):
making this a patch update feels correct to me

PR #734, opened 2022-10-10
Honour the error status code

rowanmanning:
We have useful information across all our apps which we're just ignoring. We use HTTP errors extensively which set either a `status` or `statusCode` property and there's no way for these errors to actually impact the status code which is sent to the end-user. This means that our Fastly logging isn't particularly useful, e.g. we can't say:

> Hmm, Fastly is logging a lot of `502` errors, I wonder which upstream
> service is causing us issues.

I don't think that this was ever given lots of thought as it was added 5 years ago without much of a description: 3f0f2af

Does anyone object to us honouring an error's status code if it has one?

## Why now?

This has been annoying me for _ages_, but I think I finally have a legit reason to add it in. [I'm working on surfacing proper error codes for next-topic-tracker-api](https://github.com/Financial-Times/next-topic-tracker-api/pull/165), which will now correctly throw errors with `502` and `504` status codes. Right now these will never appear in Fastly because of this n-express code, and we need them to appear in Fastly so that we can adjust our monitoring.

---

Relates to [CPREL-348](https://financialtimes.atlassian.net/browse/CPREL-348).

### DIFF
--- a/src/lib/instrument-listen.js
+++ b/src/lib/instrument-listen.js
@@ -51,9 +51,19 @@ module.exports = class InstrumentListen {
 			// Optional fallthrough error handler
 			// eslint-disable-next-line no-unused-vars
 			this.app.use((err, req, res, next) => {
+				let statusCode = err.statusCode || err.status || 500;
+
+				// There's clearly an error, so if the error has a status
+				// code of less than `400` we should default to `500` to
+				// ensure bad error handling doesn't send false positive
+				// status codes
+				if (statusCode < 400) {
+					statusCode = 500;
+				}
+
 				// The error id is attached to `res.sentry` to be returned
 				// and optionally displayed to the user for support.
-				res.statusCode = 500;
+				res.statusCode = statusCode;
 				res.end(res.sentry + '\n');
 			});
 


jkerr321:
I reaaaaaally like this! Simple and sensible 🙌

jkerr321 (approved):

src/lib/instrument-listen.js
```diff
@@ -51,9 +51,19 @@ module.exports = class InstrumentListen {
...
+				// ensure bad error handling doesn't send false positive
+				// status codes
+				if (statusCode < 400) {
+					statusCode = 500;
+				}
+
 				// The error id is attached to `res.sentry` to be returned
 				// and optionally displayed to the user for support.
-				res.statusCode = 500;
+				res.statusCode = statusCode;
```
jkerr321, line 66: Is any additional validation required? are the values or types for `err.statusCode` and `err.status` enforced anywhere upstream? If not, what are the implications of `res.statusCode` being something unexpected (i.e. not `500`-`599`)?

rowanmanning: Great question!

Types aren't enforced for either of these properties, they can technically be anything so maybe it does make sense to check this. Digging into Node.js internals [this value does get cast to an integer](https://github.com/nodejs/node/blob/5fc9d3c480bb23cd348169cd0f3c3d5c1799a287/lib/_http_server.js#L295) when the response is sent. So that means if `statusCode` is a stringified number then it'll be properly sent. I think maybe it makes more sense for us to do this ourselves though so that we can just send 500 if casting to a number doesn't work. I'll open another PR.

The other thing that happens in Node.js internals is that if the status code is invalid an [error is thrown](https://github.com/nodejs/node/blob/5fc9d3c480bb23cd348169cd0f3c3d5c1799a287/lib/_http_server.js#L297). This isn't ideal so I think we should also make sure that the status is between `400` and `599`. Again will PR this.

PR #735, opened 2022-10-10
Guard against bad error status codes

rowanmanning:
Thanks to @jkerr321 for raising this, it's totally possible for these error statuses to be something unexpected. This PR guards against any case where the status code cannot be converted to a number or isn't a valid HTTP error status code.


jkerr321 (approved):
Noice

PR #736, opened 2022-10-10
Add a withSentry option to allow disabling n-raven

rowanmanning:
This option is the first step towards us being able to provide an alternative for Sentry, or at least unpick our Sentry implementation from n-express. This is needed for us to even properly experiment in this space, we'd like to be able to trial disabling Sentry on some of our low-traffic internal apps (e.g. Houston) and it's currently not possible at all.

Notes on my choices:

  - Because importing the n-raven module has side-effects (i.e. just requiring it will register uncaught exception handlers), we had to move the `require` statements into conditionals within the n-express setup code. This is a little ugly. We could possibly abstract this into a new file within n-express if you really hate this.

  - I opted for a `withSentry` option defaulting to `true` as this is consistent with the other n-express options.

  - In production, the default error handler outputs the Sentry error ID if it's enabled and we no longer have that data if the `withSentry` option is `true`. I decided a reasonable alternative which doesn't give away any important error information would be to send the status code and status message associated with the error.

  - For now (at the request of @joelcarr) I added a warning message if you set the `withSentry` option to `false`. This is to highlight that it's still an experimental option and so that we in the Reliability team can keep track of any apps which have disabled Sentry prematurely.

### DIFF
--- a/main.js
+++ b/main.js
@@ -24,7 +24,6 @@ const consentMiddleware = require('./src/middleware/consent');
 // logging and monitoring
 const metrics = require('next-metrics');
 const serviceMetrics = require('./src/lib/service-metrics');
-const raven = require('@financial-times/n-raven');
 const nLogger = require('@financial-times/n-logger').default;
 
 // utils
@@ -54,6 +53,7 @@ const getAppContainer = (options) => {
 			withBackendAuthentication: true,
 			withFlags: false,
 			withConsent: false,
+			withSentry: true,
 			withServiceMetrics: true,
 			healthChecks: []
 		},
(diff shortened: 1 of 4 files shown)


main.js
```diff
@@ -78,13 +78,29 @@ const getAppContainer = (options) => {
...
+		// we're just deferring it until later on, when the main `express`
+		// function is called
+		const raven = require('@financial-times/n-raven');
+		app.use(raven.requestHandler());
+	} else {
+		nLogger.warn({
+			event: 'EXPERIMENTAL_OPTION',
+			message: `The 'withSentry' option is experimental and should only be set to 'false' in non-critical applications for now. This may impact our ability to debug issues with this app.`,
+			optionName: 'withSentry'
+		});
```
rowanmanning, line 102: Just so I've got somewhere to put it for now, here's the future Splunk search to find any apps which have disabled Sentry:

```
index=heroku (source="dotcom-*" OR source="next-*" OR source="cp-*") event=EXPERIMENTAL_OPTION optionName=withSentry
```

src/lib/instrument-listen.js
```diff
@@ -3,24 +3,24 @@
...
 // @ts-nocheck
 
-const raven = require('@financial-times/n-raven');
 const metrics = require('next-metrics');
 const nLogger = require('@financial-times/n-logger').default;
 const http = require('http');
 const https = require('https');
 const denodeify = require('denodeify');
 const path = require('path');
+const {STATUS_CODES} = http;
```
andygout, line 12: TIL

andygout (approved):
👍 

Feels like a minor release?

src/lib/instrument-listen.js
rowanmanning: Yep the number of times I see people including a library for this! It's right there in `http` 😁

jkerr321:
Nice Rowan - good explanations for your choices and thank you for including those in the commit message too 👍

PR #740, opened 2022-10-25
update next-metrics to 7.6

apaleslimghost:


package.json
```diff
@@ -19,7 +19,7 @@
     "express": "^4.17.3",
     "isomorphic-fetch": "^3.0.0",
     "n-health": "^8.0.1",
-    "next-metrics": "^7.5.0",
+    "next-metrics": "7.6",
```
ivomurrell, line 22: **question**: is there a reason why we're not allowing newer minor versions of `next-metrics` anymore?

ivomurrell (approved):

package.json
apaleslimghost: the reason was i forgot the `^` in the `npm install` command 😅

ivomurrell: that would have caught me out too, I'm never sure when npm will implicitly require an exact version or not 😅

PR #745, opened 2022-12-07
Correctly handle ERR_STREAM_WRITE_AFTER_END errors

rowanmanning:
We got an [app crash alert in the ft-next-support channel today](https://financialtimes.slack.com/archives/C042NBBTM/p1670374854221419) where next-syndication-api was sending a response after one has already been sent. I'm certain we need to do some work in that app too, but the error doesn't have enough detail for me to work out where the response is being written.

The error highlighted that we don't handle this case particularly well - [the Express documentation suggests](https://expressjs.com/en/guide/error-handling.html#the-default-error-handler) that you should always fall back to the default error handler if headers have already been sent to avoid this. I tracked this down to the [finalhandler](https://www.npmjs.com/package/finalhandler) module (which Express uses) and it [correctly terminates the request](https://github.com/pillarjs/finalhandler/blob/5ceb3e3e2482404cb71e9810bd10a422fe748f20/index.js#L128) instead of causing a crash.

I decided to use Reliability Kit's serializers here so that we have the same amount of detail as in our other errors. The main useful thing for debugging next-syndication-api will be having a log of the request that was made so that I can work out which route is erroring.

### DIFF
--- a/src/lib/instrument-listen.js
+++ b/src/lib/instrument-listen.js
@@ -10,6 +10,8 @@ const https = require('https');
 const denodeify = require('denodeify');
 const path = require('path');
 const {STATUS_CODES} = http;
+const serializeError = require('@dotcom-reliability-kit/serialize-error');
+const serializeRequest = require('@dotcom-reliability-kit/serialize-request');
 
 const fs = require('fs');
 const readFile = denodeify(fs.readFile);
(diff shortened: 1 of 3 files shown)


andygout (approved):
One non-blocking question: The documentation suggests that in this scenario `next()` will  go straight to the Express default error handler (and close the connection and fail the request), which implies it would circumvent any other middleware that would otherwise be called.

Do you have this same understanding?

rowanmanning:
I think they're making some assumptions in their documentation that you'd only have one error handler. So yeah this only works if it's done in the _very_ last Express error handler. The documentation's maybe a little misleading there, and If we mounted another error handling middleware after this one the `next` would send the error on to that.

Luckily I think it's impossible to add more error handling middleware after this one because of the dirty trick we use in n-express to sneak this middleware in immediately before the call to `listen`

src/lib/instrument-listen.js
```diff
@@ -77,6 +79,21 @@ module.exports = class InstrumentListen {
 				res.statusCode = isValidErrorStatus ? statusCode : 500;
```
ivomurrell, line 79: **question**: was it intentional to only fallback to the default error handler after setting the status code? the code sample in your linked express documentation shows the call to `next` happening before the code is set. I have no idea if setting the status code this late has any bearing: I guess not if the headers have already been sent?!

rowanmanning: Yeah I'm pretty sure it won't have an impact at all because HTTP sends the headers after the status code, so if headers are sent then status has been too. I wonder if I should move it just to be safe? 🤔

ivomurrell (approved):

src/lib/instrument-listen.js
ivomurrell: I'm _sure_ it's fine 😄

PR #746, opened 2022-12-07
CPP-1000: Migrate to Tool Kit

joelcarr:
Migrating `n-express` to Tool Kit. We have not migrated code coverage setup as it hasn't worked on CI. Will reintroduce code coverage when [CPP-1026](https://financialtimes.atlassian.net/browse/CPP-1026) is complete.


rowanmanning (commented):
Non-blocking but is it intentional that we stop also building/testing in Node.js 14? Or is this not something Tool Kit can do yet? I know it's end-of-life in April but we do still have a few apps running Node.js 14 and having n-express run tests on it means we can't accidentally introduce a breaking change

.circleci/config.yml
```diff
@@ -1,220 +1,72 @@
...
-  renovate-nori-build-test:
-    when:
-      not:
-        equal:
-          - scheduled_pipeline
-          - << pipeline.trigger_source >>
-    jobs:
-      - waiting-for-approval:
-          type: approval
+            - tool-kit/test
```
ivomurrell, line 64: **issue**: the `tool-kit/test` jobs' names have been appended by the Node version you're using so you need to include the full names here

You could do either:
```suggestion
            - tool-kit/test-16.14
```
which is what the migration script would have generated previously, or:
```suggestion
            - tool-kit/test-16.14
            - tool-kit/test-14.19
```
which would only publish the package if it passes for Node 14 as well. the latter makes more sense to me in retrospect though maybe there was a reason I didn't implement it like that at the time...

.editorconfig
```diff
@@ -0,0 +1,21 @@
+root = true
```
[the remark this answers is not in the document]
joelcarr: I think so! I will create a ticket.

package.json
```diff
@@ -4,11 +4,13 @@
   "main": "main.js",
   "types": "typings/n-express.d.ts",
   "scripts": {
-    "test": "make test",
+    "test": "dotcom-tool-kit test:local",
     "test:types": "tsc",
```
rowanmanning, line 8: **question (non-blocking):** would adding something like [this plugin](https://github.com/Financial-Times/dotcom-biz-ops-package-updater/tree/main/toolkit/typescript-check) to Tool Kit make sense? It does the job of `test:types` in n-express and would allow it to be done as part of `npm test` in future. I can see us needing it in a few places as we maintain a bunch of JavaScript libraries which we want to ensure are still type-safe.

ivomurrell: I just created a [PR](https://github.com/Financial-Times/dotcom-tool-kit/pull/338) to add a TypeScript plugin to Tool Kit that includes a task called [`TypeScriptTest`](https://github.com/Financial-Times/dotcom-tool-kit/blob/e79361c5b181d5901e91d2f59f05b31ff8edd5a0/plugins/typescript/src/tasks/typescript.ts#L42-L46) that should do what you describe

rowanmanning: Amazing! I'll check if this works for our case today

PR #749, opened 2022-12-15
bump n-flags-client

jkerr321:
to pull in https://github.com/Financial-Times/n-flags-client/pull/179 which updates logs to check for headers before calling the header.entries() method


joelcarr (approved):

rowanmanning (approved):
Nice 👍

PR #760, opened 2023-03-31
BREAKING: Default to switching off Sentry

rowanmanning:
This changes the `withSentry` option to default to `false`, so apps need to explicitly opt into Sentry logging. The next planned major version of n-express will remove this option entirely so I've also added a deprecation warning so that we can see how many apps still use it.

This should not be merged until we're ready to roll out the mid-April breaking changes.


joelcarr (approved):
PR of the Sentry…

PR #771, opened 2023-05-09
build: don't wait for approval for tag builds

apaleslimghost:


.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ workflows:
             branches:
               only: /(^renovate-.*|^nori/.*)/
             tags:
-              only: /^v\d+\.\d+\.\d+(-.+)?/
+              ignore: /.*/
```
ivomurrell, line 32: **nitpick**: we can remove the `tags` stanza completely as all tags are ignored by default iiuc

ivomurrell (approved):
**nitpick**: the `ci` git commit type would be more appropriate 😄

Two v minor nitpicks, merge if you have any lack of inclination to resolving either of them!

PR #811, opened 2024-02-23
[CPREL-984] Remove backend auth loophole

alexmuller:
For many years this conditional has meant that an app called next-foo will bypass backend authentication for routes that begin /foo.

This is confusing and unnecessary and we should remove it. It'll be a breaking change.

There are very few apps where the first part of their path matches the app name and they all have backend authentication set up correctly:

```
"^/article"                 next-article
"^/globetrotter/(.+)"       next-globetrotter
"^/newsletter-signup"       next-newsletter-signup
"^/profile"                 next-profile
"^/static"                  next-static
```

[JIRA ticket](https://financialtimes.atlassian.net/browse/CPREL-984)

I also added a little type fix because I don't get to write code very much these days and it's fun 🤷

### DIFF
--- a/main.js
+++ b/main.js
@@ -144,7 +144,7 @@ const getAppContainer = (options) => {
 	);
 
 	if (options.withBackendAuthentication) {
-		backendAuthentication(app, meta.name);
+		backendAuthentication(app);
 	}
 
 	// feature flags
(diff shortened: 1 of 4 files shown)


ivomurrell (approved):

rowanmanning (approved):
Lovely 👍

apaleslimghost (approved):
qu'est-ce que le fuck

CyntiBinti (approved):

PR #822, opened 2024-05-13
Remove the graphite checks

rowanmanning:
This is a breaking change. We're removing the various health check types that rely on reading Graphite metrics because, as of Q3 2024, we will no longer be able to read this data.

For now we recommend removing these checks from your apps. The alerts aren't very commonly-used and we have better ways of monitoring uptime now. [We have a full migration guide on Confluence](https://financialtimes.atlassian.net/wiki/spaces/DS/pages/8408989698/Migrating+an+app+away+from+Graphite-based+health+checks).

---

See-Also: [CPREL-1056](https://financialtimes.atlassian.net/browse/CPREL-1056)

[CPREL-1056]: https://financialtimes.atlassian.net/browse/CPREL-1056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ivomurrell (approved):
**nitpick**: we should also remove a mention of the graphite name from the [README](https://github.com/Financial-Times/n-express?tab=readme-ov-file#optional)

PR #831, opened 2024-06-14
feat: allow Node 22

alexmuller:
People can use this now if they want to.

https://github.com/nodejs/release


rowanmanning (approved):

joelcarr:
What should our Node version policy be? Should packages support the current version as well or should we limit it to LTS versions?

alexmuller:
Not 100% sure on this but my gut feeling is that we should test on and support the latest we realistically can. That way by the time people come to upgrade their systems they hopefully won't get that many npm warnings because they'll already be on a package version with support.

But @rowanmanning has thought more about this than me.

alexmuller:
Sorry I should have added the context for this change - Ads tried to see if they could jump straight to 22 and it was fine except for this alert. So let's not hold them back if they want to be cutting edge testers for us.

PR #836, opened 2024-07-29
Remove the Service Metrics alert

rowanmanning:
This removes the annoying alert when a service is not recognised. This causes a lot of churn and there's no need for it. next-metrics now guesses the service name based on the request URL's host name (https://github.com/Financial-Times/next-metrics/pull/583) and so this alert will never fire.

This also allows us to do a lot of simplification because we don't have to handle undefined services.

For all the removed things (e.g. types/options):
1. these was never part of the public API for n-express
2. we don't seem to be using them anywhere

---

See-also: [CPREL-1123](https://financialtimes.atlassian.net/browse/CPREL-1123)

[CPREL-1123]: https://financialtimes.atlassian.net/browse/CPREL-1123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apaleslimghost (approved):
bye felicia

PR #838, opened 2024-07-31
only include isomorphic-fetch if the application has explicitly disabled native fetch

apaleslimghost:
this will allow us to gradually enable native fetch in apps without a breaking change. by default, apps on Node 18 using Tool Kit set this flag. once everything has migrated to native fetch we can remove the dependency on `isomorphic-fetch`.

### DIFF
--- a/main.js
+++ b/main.js
@@ -6,7 +6,13 @@
  * @typedef {import("./typings/metrics").TickingMetric} TickingMetric
  */
 
-require('isomorphic-fetch');
+// only include isomorphic-fetch if the application has explicitly disabled native fetch
+if (
+	process.allowedNodeEnvironmentFlags.has('--no-experimental-fetch') &&
+	process.execArgv.includes('--no-experimental-fetch')
+) {
+	require('isomorphic-fetch');
+}
 
 const fs = require('fs');
 const path = require('path');
@@ -68,7 +74,8 @@ const getAppContainer = (options) => {
 	if (options.withAb) {
 		logger.warn({
 			event: 'WITHAB_OPTION_DEPRECATED',
-			message: 'The \'withAb\' option is deprecated and no longer supported by n-express or n-flags-client'
+			message:
+				"The 'withAb' option is deprecated and no longer supported by n-express or n-flags-client"
 		});
 	}
 


main.js
```diff
@@ -6,7 +6,13 @@
  * @typedef {import("./typings/metrics").TickingMetric} TickingMetric
  */
 
-require('isomorphic-fetch');
+// only include isomorphic-fetch if the application has explicitly disabled native fetch
+if (
+	process.allowedNodeEnvironmentFlags.has('--no-experimental-fetch') &&
+	process.execArgv.includes('--no-experimental-fetch')
```
ivomurrell, line 12: **nitpick**: not that we support said versions anymore, but `process.allowedNodeEnvironmentFlags.has('--no-experimental-fetch')` resolving to `false` implies that this is an older Node version without support for native fetch so we should fallback to `isomorphic-fetch` in that case

```suggestion
	!process.allowedNodeEnvironmentFlags.has('--no-experimental-fetch') ||
	process.execArgv.includes('--no-experimental-fetch')
```

main.js
```diff
@@ -6,7 +6,13 @@
  * @typedef {import("./typings/metrics").TickingMetric} TickingMetric
  */
 
-require('isomorphic-fetch');
+// only include isomorphic-fetch if the application has explicitly disabled native fetch
```
ivomurrell, line 9: **suggestion**: could we expand this comment to add a bit of context on why we're doing this? afaict it's to satisfy problem hub as there won't be any behavioural change (as `isomorphic-fetch` already checks if native fetch is present)

ivomurrell (approved):

main.js
apaleslimghost: oh very good point

main.js
```diff
@@ -68,7 +74,8 @@ const getAppContainer = (options) => {
 	if (options.withAb) {
 		logger.warn({
 			event: 'WITHAB_OPTION_DEPRECATED',
-			message: 'The \'withAb\' option is deprecated and no longer supported by n-express or n-flags-client'
+			message:
+				"The 'withAb' option is deprecated and no longer supported by n-express or n-flags-client"
```
rowanmanning, line 78: This looks like prettier or something edited an unrelated bit of code. The config seems wrong because ESLint is complaining about double quotes.

ivomurrell: bike shedding territory:
- i think it's fine to include a prettier change in an unrelated commit, nbd imo
- eslint is wrong in this case, prettier [chooses double quotes here because it's more readable](https://prettier.io/docs/en/rationale#strings) and eslint shouldn't be complaining about that. should probably be using the [eslint prettier plugin](https://github.com/prettier/eslint-config-prettier) to avoid these conflicts.

main.js
apaleslimghost: so this idea is [from Jen's original native fetch investigation](https://financialtimes.atlassian.net/wiki/spaces/CPP/pages/8042020974/Node.js+18+and+Native+Fetch#Moving-to-future-state---one-option). which i should have read a bit closer, because the reason she's suggesting doing it this way is so we can add a warning here. i'll do that.

rowanmanning:
Gonna close this as #943 removed isomorphic-fetch entirely, released in [v32](https://github.com/Financial-Times/n-express/releases/tag/v32.0.0)

PR #841, opened 2024-08-21
Remove the Node.js version healthcheck

rowanmanning:
There's no need for this to be an operational alert. This is covered well by several Problem Hub problems and I prefer the code we don't have to maintain.


CyntiBinti (approved):
🙌🏽  less (unnecessary) checks

PR #842, opened 2024-08-21
Remove all default healthchecks

rowanmanning:
This is a change that just kept on giving!

So I started out wanting to remove the health check that verifies whether next-metrics is configured properly. We don't need this any more, I couldn't care less if an app stops sending metrics 🙈 we will rethink this in future, the right way is probably a heartbeat metric in OpenTelemetry.

Anyway, once I removed the metrics health check there were no more default checks. This means there are no more ticking checks at all as far as I can tell. This allowed me to remove a bunch of redundant code from the internals.

I tested this by symlinking it into several different n-express apps and everything starts up fine, including for apps that define their own healthchecks.


main.js
```diff
@@ -105,7 +104,7 @@ const getAppContainer = (options) => {
 	app.use(vary);
 
 	if (!options.demo) {
-		instrumentListen.addMetrics(healthChecks(app, options, meta));
+		setupHealthEndpoint(app, options, meta);
```
apaleslimghost, line 107: **suggestion**: this is now just creating a route, and it's the only thing creating a route that's in another file. i think it would be easier to follow if we inlined it.

apaleslimghost (approved):
👀 👀 👀 👀 👀 

**chore**: we can remove the `TickingMetric` typedef from `metrics.d.ts`. it's [not used anywhere outside of `n-express`](https://github.com/search?q=org%3AFinancial-Times+TickingMetric+-repo%3Afinancial-times%2Fn-express&type=code).

main.js
rowanmanning: I agree, was just trying to keep the diff small and readable because a lot is already changing 😁 I'll move it as a separate commit

rowanmanning: Actually having messed around with this it makes the diff quite messy and I'm not sure if I want my name against it all 😄 is it alright to reconsider the structure separately? There's quite a lot of stuff that could move around

PR #864, opened 2024-10-17
Sort out the types

rowanmanning:
This does a bunch of stuff to bring the types up to date:

  * Updates the TypeScript-related packages to match the versions of Node.js we support (18+ currently)
  * Fixes all the TypeScript errors, all of these turned out to be fine and non-breaking
  * Adds the Tool Kit TypeScript package to the test command so that things can't break again
  * Adds strict type checking to avoid a few gotchas being added later
  * Migrates JSDoc to `@import` statements which are neater


apaleslimghost (approved):
nice! these definitions were added by another team so they've not had much love, thanks for sorting them out!

PR #884, opened 2024-11-04
chore!: stop sending metrics

rowanmanning:
This doesn't fully remove next-metrics to avoid a large migration, however it does stop n-express from sending any metrics to Graphite by default. This is a breaking change because the lack of metrics could trigger alerts for some apps as data will stop being available for things like `fetch`.

The second part of this work will arrive when we roll out next-metrics v13, after [this PR](https://github.com/Financial-Times/next-metrics/pull/603) is merged.

**Note: this is a breaking change, note that it's being PRed into the `next` branch**


joelcarr (approved):
💯 Lets go!

PR #940, opened 2025-03-19
CPP-2398: add subscription middleware

meel-io:
## Description

Extracts the subscription details from the `ft-user-subscription` header, parses it and set the local property `subscription`.

The `ft-user-subscription` header is expected to have this format: 
```
ft-user-subscriptions:status=subscribed;productCodes=P1;access=isB2c,isStaff;licenceIds=008ce617-4453-438a-9760-7c3f3cd3b67f
```
as recently implemented on the [Preflight task](https://github.com/Financial-Times/next-preflight/blob/main/server/tasks/membership/subscription.js#L84-L112).

The anon middleware will be able to read this property to determine the `isUserSubscribed` property, instead of reading from the `ft-user-subscription-status` header, which is set by Ammit API. Although we're expecting apps to migrate to the new subscription object to read the user subscription status.

### DIFF
--- a/main.js
+++ b/main.js
@@ -32,6 +32,7 @@ const robots = require('./src/middleware/robots');
 const security = require('./src/middleware/security');
 const vary = require('./src/middleware/vary');
 const logVary = require('./src/middleware/log-vary');
+const subscription = require('./src/middleware/subscription');
 const anon = require('./src/middleware/anon');
 const teapot = fs.readFileSync(
 	path.join(__dirname, 'src/assets/teapot.ascii'),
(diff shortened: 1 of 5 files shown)


main.js
```diff
@@ -156,6 +157,10 @@ const getAppContainer = (options) => {
 		app.use(anon.middleware);
 	}
 
+	if (options.withSubscriptionDetails) {
+		app.use(subscription.middleware);
```
ivomurrell, line 161: **issue**: shouldn't the middleware be mounted before the `anon` middleware if you plan to reference the properties it sets?

rowanmanning:
Not blocking this, it's our bad for not really documenting anywhere, but @apaleslimghost and I have been trying to reduce the footprint of n-express for quite some time and this informed decisions like keeping various Reliability Kit middleware out of this repo. The intention is that each team should be able to make decisions about what's included by mounting Express middleware themselves.

Is there any reason this can't be a separate module? I don't know the background on this work so there may be a legit reason why it _has_ to be in n-express.

PR #944, opened 2025-03-20
CPP-2397: Derive isUserSubscribed property from ft-user-subscription header

meel-io:
## Description

Extracts the user subscription status from the `ft-user-subscription` header and uses it to set the `isUserSubscribed` property. 

The `ft-user-subscription` header [should always be set](https://github.com/Financial-Times/ft.com-cdn/blob/3ef5c860e390fe882e51a2a63134566cf53b4112/src/vcl/next-preflight.vcl#L67) for apps behind the FT.com CDN. However, we’re considering cases where it might be missing or have an incorrect value. In those scenarios, we log a warning and fall back to the previous `ft-user-subscription-status` header. This will allow us to detect issues with the new header before removing the old one.

The `ft-user-subscription-status` is currently being set by Ammit API, our deprecated flags system. The new `ft-user-subscription` header contains the user subscription details without relying on Ammit API flags, allowing us to decouple FT.com features that depend on the `ft-user-subscription-status` header or other subscription data from Ammit API.

### DIFF
--- a/src/middleware/anon.js
+++ b/src/middleware/anon.js
@@ -1,3 +1,5 @@
+const logger = require('@dotcom-reliability-kit/logger');
+
 /**
  * @import {Callback, Request, Response} from '../../typings/n-express'
  */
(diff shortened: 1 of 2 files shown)


src/middleware/anon.js
```diff
@@ -14,10 +16,53 @@ function AnonymousModel (req) {
 		this.userIsLoggedIn = true;
 		this.userIsAnonymous = false;
 		// set by ammit task in preflight
```
ivomurrell, line 18: **nitpick**: we're now using the `status` property set by the subscription task

```suggestion
		// set by subscription task in preflight
```

src/middleware/anon.js
```diff
@@ -14,10 +16,53 @@ function AnonymousModel (req) {
...
-		this.userIsSubscribed = req.get('ft-user-subscription-status') === 'subscribed';
+		this.userIsSubscribed = getSubscriptionStatus(req) === 'subscribed';
 	}
 }
 
+/**
+ * @param {Request} req
+ */
+function getSubscriptionStatus (req) {
+	// e.g.: status=subscribed;productCodes=P2,MPR;licenceIds=00000000-0000-0000-0000-000000000000;access=isB2c,isStaff
```
ivomurrell, line 27: **praise**: very useful comments

ivomurrell (approved):

CyntiBinti (approved):

PR #959, opened 2025-04-03
feat: remove default error middleware

apaleslimghost:
a) everything this middleware does, the default express error handler does (we used to do Raven/Sentry error logging here)
b) also, @dotcom-reliability-kit/middleware-render-error-info

### DIFF
--- a/src/lib/instrument-listen.js
+++ b/src/lib/instrument-listen.js
@@ -5,8 +5,6 @@ const http = require('http');
 const https = require('https');
 const path = require('path');
 const {readFile} = require('fs/promises');
-const {STATUS_CODES} = http;
-const serializeRequest = require('@dotcom-reliability-kit/serialize-request');
 
 module.exports = class InstrumentListen {
 	constructor (app, meta, initPromises) {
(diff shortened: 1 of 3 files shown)


src/lib/instrument-listen.js
```diff
@@ -5,8 +5,6 @@ const http = require('http');
 const https = require('https');
 const path = require('path');
 const {readFile} = require('fs/promises');
-const {STATUS_CODES} = http;
-const serializeRequest = require('@dotcom-reliability-kit/serialize-request');
```
rowanmanning, line 9: **suggestion:** this should be uninstalled as well, it's the only use

rowanmanning (changes requested):
Tested locally, confirmed the behaviour is the same 👍

src/lib/instrument-listen.js
apaleslimghost: oh i can't read

apaleslimghost: that's `serialize-request`, i was looking for `serialize-error` lol

rowanmanning (approved):

PR #962, opened 2025-04-03
feat!: support express v5

apaleslimghost:
this is a breaking change, as apps currently aren't explicitly depending on any version of Express, and so would be switching to v5 when updating to a version of `n-express` with this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "next-metrics": "^13.0.0"
   },
   "peerDependencies": {
-    "express": "^4.21.2"
+    "express": "^4.21.2 || 5.x"
   },
   "devDependencies": {
     "@dotcom-tool-kit/component": "^5.1.13",
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "node": "20.x || 22.x"
       },
       "peerDependencies": {
-        "express": "^4.21.2"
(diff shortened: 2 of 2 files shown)


kavanagh:
Ooh I had the exact same idea and found Kara beat me to it. I support this feature.

package.json
```diff
@@ -18,7 +18,7 @@
     "next-metrics": "^13.0.0"
   },
   "peerDependencies": {
-    "express": "^4.21.2"
+    "express": ">=4.21.2 <5.0.0 || 5.x"
```
ivomurrell, line 21: **nitpick**: using the caret range as previous would be equivalent and simpler imo

```suggestion
    "express": "^4.21.2 || 5.x"
```

ivomurrell (approved):

package.json
apaleslimghost: why did i not think of that 🤔

rowanmanning: **question:** why the `5.x` instead of `^5.1.0`? I can't remember the exact difference

ivomurrell: generally we use `X.x` in peer dependencies to represent as wide a range as possible as they can be quite disruptive otherwise. technically you could do `^X.0.0` instead but I think `X.x` is clearer

PR #978, opened 2025-05-01
Simplify the `robots.txt` middleware

rowanmanning:
We questioned whether this was useful as part of our n-express audit: https://financialtimes.atlassian.net/wiki/spaces/CPP/database/8965226525

We serve a full `robots.txt` on FT.com via the CDN so we wondered whether this was redundant, however it's important that we don't allow crawlers to index systems that are accessed directly and do not use a backend key.

We need to keep this, but I think we can simplify a little:

  * I removed the "Allow Twitterbot" rule - The original commit here 4bc78b9883dfe20116827a9a76758ae8b465a2ef does not explain why it was added, if this causes an issue then we can rethink but at the moment I can't imagine a use-case for us wanting Twitter link previews for any of our apps direct from the Heroku/Hako app URLs.

  * I fixed the cache-control header (`max-age` requires an equals and not a colon) added in 7a3ed3ca72299c5eabd156a9bc519c0276f284eb. I think this never worked. I also increased the cache time from one hour to 30 days because we never change this.

  * I moved directives from a file into the middleware itself as a string. I don't want to encourage editing this and we can save some admittedly small sync file operations during startup. Every little helps eh?

### DIFF
--- a/src/middleware/robots.js
+++ b/src/middleware/robots.js
@@ -2,19 +2,18 @@
  * @import {Callback} from '../../typings/n-express'
  */
 
-const fs = require('fs');
-const path = require('path');
-const robots = fs.readFileSync(path.join(__dirname, '../assets/robots.txt'), {
-	encoding: 'utf8'
-});
+const headers = {
+	'Content-Type': 'text/plain',
+	'Cache-Control': `public, max-age=${60 * 60 * 24 * 30}` // One month
+};
 
-/**
- * @type {Callback}
- */
-module.exports = (_req, res) => {
-	res.set({
-		'Content-Type': 'text/plain',
-		'Cache-Control': 'max-age:3600, public'
-	});
-	res.send(robots);
+// This is a string rather than a separate file because we don't want
+// this fallback robots.txt file to be modified or added to. This is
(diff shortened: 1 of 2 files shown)


src/middleware/robots.js
```diff
@@ -2,19 +2,18 @@
...
-/**
- * @type {Callback}
- */
-module.exports = (_req, res) => {
-	res.set({
-		'Content-Type': 'text/plain',
-		'Cache-Control': 'max-age:3600, public'
-	});
-	res.send(robots);
+// This is a string rather than a separate file because we don't want
```
olusoladavid, line 10: **non-blocking thought:** maybe it is useful to add an assertion about the expected response of the `/robots.txt` path in the `robots.txt` test case

olusoladavid (approved):
Thanks for the audit

CyntiBinti (approved):

src/middleware/robots.js
rowanmanning: Makes sense 👍 haven't discussed with kara/ivo yet but I'm tempted to move this to use `node --test` at some point and revise the tests that we have, will add a robots test separately

PR #979, opened 2025-05-01
Remove the via middleware

rowanmanning:
This was added by me as an experiment. We didn't end up using it, it's not documented, and it's not used anywhere that I can find.


CyntiBinti (approved):

olusoladavid (approved):
Hyrum's law suggests this is still a breaking change 😄

PR #1005, opened 2025-06-25
Skip anon middleware for non-Preflight requests

meel-io:
## Description

Releasing the current version of this middleware caused some apps to unexpectetly log a [high volume](https://financialtimes.atlassian.net/browse/CPP-2551) of `SUBSCRIPTION_HEADER_MISSING` warning logs, concretely on [next-control-centre](https://github.com/Financial-Times/next-control-centre). We believe that these might be coming from requests that don't go through Preflight, so therefore don't get a a valid `ft-user-subscription` header.  
 
To mitigate this, we're skipping this middleware logic on requests that are with prefixed with  `__`. These include assets, healthchecks, etc. 

We're also skipping `POST` and `PUT` requests as these also [skip Preflight](https://github.com/Financial-Times/ft.com-cdn/blob/dafd68b65a27b90352e34f0b00b38b0decbf86b8/src/vcl/next-preflight.vcl#L99-L102).

Releasing this version on `next-control-centre` and checking if the it doesn't log a high volume of these warnings should confirm if this fix worked as expected.

### DIFF
--- a/src/middleware/anon.js
+++ b/src/middleware/anon.js
@@ -88,6 +88,19 @@ function showFirstClickFree (req, res) {
  * @type {Callback}
  */
 function anonymousMiddleware (req, res, next) {
+	// Skip for static assets, healthchecks, etc.
+	// These routes shouldn't need the anonymous status
+	if (req.path.indexOf('/__') === 0) {
+		return next();
+	}
+
+	// Skip for POST, PUT requests
+	// These routes skip preflight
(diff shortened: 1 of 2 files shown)


olusoladavid (approved):
Thanks for the fix. Is the plan to backport to v31?

src/middleware/anon.js
```diff
@@ -88,6 +88,19 @@ function showFirstClickFree (req, res) {
  * @type {Callback}
  */
 function anonymousMiddleware (req, res, next) {
+	// Skip for static assets, healthchecks, etc.
```
olusoladavid, line 91: **question:** do we just want to check if preflight has been skipped and not duplicate the logic here?

